PR #2, opened 2020-02-27
Pipe all output from generic output to stderr

sirdarckcat:
This is needed because `make ip` expects a clean stdout

### DIFF
--- a/samples/kctf-conf/base/Makefile
+++ b/samples/kctf-conf/base/Makefile
@@ -279,8 +279,8 @@ ${CLUSTER_GEN}/remote-healthcheck-image: ${CLUSTER_GEN}/healthcheck-image-tagged
 	@  echo 'error: cluster config not loaded. Run kctf-config-create or kctf-config-load'
 	@  exit 1
 	@fi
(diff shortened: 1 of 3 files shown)


samples/kctf-conf/base/Makefile
```diff
@@ -279,8 +279,8 @@ ${CLUSTER_GEN}/remote-healthcheck-image: ${CLUSTER_GEN}/healthcheck-image-tagged
 	@  echo 'error: cluster config not loaded. Run kctf-config-create or kctf-config-load'
 	@  exit 1
 	@fi
-	kubectl config use-context "kctf_${PROJECT}_${ZONE}_${CLUSTER_NAME}"
-	mkdir -p ${CLUSTER_GEN}
+	kubectl config use-context "kctf_${PROJECT}_${ZONE}_${CLUSTER_NAME}" >&2
+	mkdir -p ${CLUSTER_GEN} >&2
```
sroettger: can we just do > /dev/null instead? The stdout of those commands is not interesting anyway and writing to stderr should IMO only be done for errors and warnings. I.e. I had a friend who would color stderr output in red which would make this weird.

PR #22, opened 2020-03-03
Create a base image for the healthcheck

sroettger:


base/healthcheck-docker/.dockerignore
```diff
@@ -0,0 +1,2 @@
+.gen
+Makefile
```
sirdarckcat: why do we ignore the makefile?

base/healthcheck-docker/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+# limitations under the License.
+FROM ubuntu:19.10
+
+RUN apt-get update && apt-get upgrade -y
+
+RUN /usr/sbin/useradd --no-create-home -u 1000 user
+
+RUN apt-get -y install python2.7 python-pip
+RUN pip install pwntools
+RUN mkdir -p /home/user/.pwntools-cache && echo never > /home/user/.pwntools-cache/update
```
sirdarckcat: why don't we cache this? it's the slowest part probably.

we could have this with the other kctf-nsjail-* images.

base/healthcheck-docker/files/decrypt_exploit.sh
```diff
@@ -0,0 +1,32 @@
...
+
+# decrypt
+KEY="$(cat /keys/exploit.key)"
+openssl aes-256-cbc \
+  -d \
+  -in exploit.cpio.enc \
+  -out exploit.cpio \
+  -K "${KEY}" \
+  -nosalt \
+  -iv 00000000000000000000000000000000
```
sirdarckcat: this totally should be instead  `13333333333333333333333333333337`

base/healthcheck-docker/.dockerignore
sroettger: because it's not needed for the docker build so no reason to send it to the docker daemon

base/healthcheck-docker/Dockerfile
sroettger: hmm, I would need to know where the files get installed by pip. Maybe I can use a virtualenv, wdyt?

sirdarckcat: yea virtualenv would be awesome :)

base/healthcheck-docker/.dockerignore
sirdarckcat: hmm ok

base/healthcheck-docker/Dockerfile
sroettger: done

PR #23, opened 2020-03-03
Save configuration to the active config

sirdarckcat:
Before this, the config was being set on the default configuration, which did nothing.

It's unclear to me why this worked at all.


scripts/lib/config.sh
```diff
@@ -81,10 +81,10 @@ function create_gcloud_config {
   export CLOUDSDK_ACTIVE_CONFIG_NAME="${CONFIG_NAME}"
```
sroettger: this already sets the active config

sirdarckcat: I'll debug today, but this wasn't working for me until I made this change.

sirdarckcat:
cant repro. will reopen if it makes sense later on to me

PR #36, opened 2020-03-08
Size and speed docker improvements

sirdarckcat:
Following best practices from https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run and tips from https://ubuntu.com/blog/we-reduced-our-docker-images-by-60-with-no-install-recommends so we install less packages.

For the preinstalled packages, following advice from https://www.dajobe.org/blog/2015/04/18/making-debian-docker-images-smaller/ so these are smaller.

Inspired by https://github.com/google/google-ctf/pull/48

### DIFF
--- a/base/challenge-skeleton/healthcheck/Dockerfile
+++ b/base/challenge-skeleton/healthcheck/Dockerfile
@@ -17,9 +17,6 @@ FROM kctf-healthcheck
 # This is out of caution in case that we leak the 
 COPY .gen/exploit.cpio.enc /
 
-# Needed by the webserver healthz.py
-RUN apt-get -y install python2.7
-
 COPY files /home/user
 
 CMD decrypt_exploit.sh && setpriv --init-groups --reuid user --regid user --inh-caps=-all -- /exploit/run.sh & /home/user/healthz.py
--- a/base/healthcheck-docker/Dockerfile
+++ b/base/healthcheck-docker/Dockerfile
@@ -14,14 +14,12 @@
 FROM gcr.io/kctf-docker/kctf-pwntools:latest AS pwntools
 FROM ubuntu:19.10
 
-RUN apt-get update && apt-get upgrade -y
+# needed by decrypt_exploit.sh
+RUN apt-get update && apt-get -yq --no-install-recommends install cpio openssl python2.7 && rm -rf /var/lib/apt/lists/*
 
 RUN /usr/sbin/useradd --no-create-home -u 1000 user
 
 COPY --from=pwntools /venv /venv
 RUN mkdir -p /home/user/.pwntools-cache && echo never > /home/user/.pwntools-cache/update
-
-# needed by decrypt_exploit.sh
-RUN apt-get -y install cpio openssl
 
 COPY files/decrypt_exploit.sh /usr/bin/
--- a/base/nsjail-docker/Dockerfile
+++ b/base/nsjail-docker/Dockerfile
@@ -15,8 +15,9 @@ FROM gcr.io/kctf-docker/kctf-nsjail-chroot:latest AS chroot
 FROM gcr.io/kctf-docker/kctf-nsjail-bin:latest AS bin
 FROM ubuntu:19.10
 
-RUN apt-get update && apt-get install -y uidmap libprotobuf17 libnl-route-3-200
-RUN apt-get upgrade -y
+RUN apt-get update \
+    && apt-get install -yq --no-install-recommends uidmap libprotobuf17 libnl-route-3-200 wget netcat ca-certificates \
+    && rm -rf /var/lib/apt/lists/*
 
 RUN /usr/sbin/useradd --no-create-home -u 1000 user
 
--- a/config/docker/chroot.Dockerfile
+++ b/config/docker/chroot.Dockerfile
@@ -17,5 +17,6 @@ RUN apt-get update \
     && apt-get install -yq --no-install-recommends debootstrap \
     && rm -rf /var/lib/apt/lists/* \
     && debootstrap --variant minbase --include python3 eoan /chroot \
+    && cp /etc/resolv.conf /chroot/etc/resolv.conf \
     && chroot /chroot /usr/sbin/useradd --no-create-home -u 1000 user \
     && apt-get remove --purge -y debootstrap $(apt-mark showauto)
--- a/config/docker/chroot.Dockerfile
+++ b/config/docker/chroot.Dockerfile
@@ -13,8 +13,7 @@
 # limitations under the License.
 FROM ubuntu:19.10
 
-RUN apt-get update && apt-get install -y debootstrap
-RUN apt-get upgrade -y
+RUN apt-get update && apt-get install -y --no-install-recommends debootstrap
 
 RUN debootstrap --variant minbase --include python3 eoan /chroot
 RUN chroot /chroot /usr/sbin/useradd --no-create-home -u 1000 user
--- a/config/docker/chroot.Dockerfile
+++ b/config/docker/chroot.Dockerfile
@@ -13,8 +13,9 @@
 # limitations under the License.
 FROM ubuntu:19.10
 
-RUN apt-get update && apt-get install -y --no-install-recommends debootstrap
-
-RUN debootstrap --variant minbase --include python3 eoan /chroot
-RUN chroot /chroot /usr/sbin/useradd --no-create-home -u 1000 user
-RUN /usr/sbin/useradd --no-create-home -u 1000 user
+RUN apt-get update \
+    && apt-get install -yq --no-install-recommends debootstrap \
+    && rm -rf /var/lib/apt/lists/* \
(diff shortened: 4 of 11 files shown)


sirdarckcat:
- kctf-nsjail-bin: [before](https://console.cloud.google.com/gcr/images/kctf-docker/GLOBAL/kctf-nsjail-bin@sha256:dba2827fdbbe3e5b3dfe3f07e400ee005f3efa4e1aa8c33d1b2479926cddb0cc/details?tab=info) was 164.5 MB and [after](https://console.cloud.google.com/gcr/images/kctf-docker/GLOBAL/kctf-nsjail-bin@sha256:c5a961b293781a4eee50938212415d25ae68ca8c287de9feb2c28ed4e3d1cea1/details?tab=info) is 32.5MB
- kctf-nsjail-chroot: [before](https://console.cloud.google.com/gcr/images/kctf-docker/GLOBAL/kctf-nsjail-chroot@sha256:bd5a596ab1f39134b7781c0b42f5ff7917b41cb943c2458f90df2a510b16b2cb/details?tab=info) was 131.5 MB and [after](https://console.cloud.google.com/gcr/images/kctf-docker/GLOBAL/kctf-nsjail-chroot@sha256:69f359a5d6a28db0d30c596389838e32ccf05213ab359fbe0455545b17f7fefd/details?tab=info) is 100.2 MB
- kctf-pwntools [before](https://console.cloud.google.com/gcr/images/kctf-docker/GLOBAL/kctf-pwntools@sha256:20ca51501833812f2b117e1419615332ceb961d2da579e765adf906e8942fa2e/details?tab=info) was 214 MB and [after](https://console.cloud.google.com/gcr/images/kctf-docker/GLOBAL/kctf-pwntools@sha256:e5361cbee89d958e6bf60bebb29d066bb456126bc682f381edfe791f096ca74a/details?tab=info) is 97.7 MB

base/challenge-skeleton/healthcheck/Dockerfile
```diff
@@ -17,9 +17,6 @@ FROM kctf-healthcheck
 # This is out of caution in case that we leak the 
 COPY .gen/exploit.cpio.enc /
 
-# Needed by the webserver healthz.py
-RUN apt-get -y install python2.7
```
sirdarckcat: this now is installed on `base/healthcheck-docker/Dockerfile` since it's necessary for pwntools anyway

base/healthcheck-docker/Dockerfile
```diff
@@ -14,14 +14,12 @@
 FROM gcr.io/kctf-docker/kctf-pwntools:latest AS pwntools
 FROM ubuntu:19.10
 
-RUN apt-get update && apt-get upgrade -y
```
sirdarckcat: removing upgrade commands, and instead using `apt-get update && apt-get install` so that we always install latest.

base/healthcheck-docker/Dockerfile
```diff
@@ -14,14 +14,12 @@
...
+# needed by decrypt_exploit.sh
+RUN apt-get update && apt-get -yq --no-install-recommends install cpio openssl python2.7 && rm -rf /var/lib/apt/lists/*
 
 RUN /usr/sbin/useradd --no-create-home -u 1000 user
 
 COPY --from=pwntools /venv /venv
 RUN mkdir -p /home/user/.pwntools-cache && echo never > /home/user/.pwntools-cache/update
 
-# needed by decrypt_exploit.sh
```
sirdarckcat: moved above to reduce the number of times we need to use `apt-get update`

base/nsjail-docker/Dockerfile
```diff
@@ -15,8 +15,9 @@ FROM gcr.io/kctf-docker/kctf-nsjail-chroot:latest AS chroot
 FROM gcr.io/kctf-docker/kctf-nsjail-bin:latest AS bin
 FROM ubuntu:19.10
 
-RUN apt-get update && apt-get install -y uidmap libprotobuf17 libnl-route-3-200
-RUN apt-get upgrade -y
+RUN apt-get update \
+    && apt-get install -yq --no-install-recommends uidmap libprotobuf17 libnl-route-3-200 wget netcat ca-certificates \
```
sirdarckcat: added wget, netcat, and ca-certificates since they are small packages and make debugging a lot easier

config/docker/chroot.Dockerfile
```diff
@@ -13,9 +13,10 @@
...
-RUN apt-get upgrade -y
-
-RUN debootstrap --variant minbase --include python3 eoan /chroot
-RUN chroot /chroot /usr/sbin/useradd --no-create-home -u 1000 user
-RUN /usr/sbin/useradd --no-create-home -u 1000 user
+RUN apt-get update \
+    && apt-get install -yq --no-install-recommends debootstrap \
+    && rm -rf /var/lib/apt/lists/* \
+    && debootstrap --variant minbase --include python3 eoan /chroot \
+    && echo nameserver 8.8.8.8 > /chroot/etc/resolv.conf \
```
sirdarckcat: by default, debootstrap copies the /etc/resolv.conf which is wrong! because we build this on GitHub Actions, which run in Azure, which has a weird DNS setup

base/challenge-skeleton/healthcheck/Dockerfile
sroettger: Hmm, pwntools is using the python from the venv and not the global one, that's why this should be installed here IMO.

base/nsjail-docker/Dockerfile
```diff
@@ -15,8 +15,9 @@ FROM gcr.io/kctf-docker/kctf-nsjail-chroot:latest AS chroot
 FROM gcr.io/kctf-docker/kctf-nsjail-bin:latest AS bin
 FROM ubuntu:19.10
 
-RUN apt-get update && apt-get install -y uidmap libprotobuf17 libnl-route-3-200
-RUN apt-get upgrade -y
```
sroettger: why no apt-get upgrade anymore? How do we get security fixes?

config/docker/chroot.Dockerfile
```diff
@@ -13,9 +13,10 @@
...
-RUN apt-get update && apt-get install -y debootstrap
-RUN apt-get upgrade -y
-
-RUN debootstrap --variant minbase --include python3 eoan /chroot
-RUN chroot /chroot /usr/sbin/useradd --no-create-home -u 1000 user
-RUN /usr/sbin/useradd --no-create-home -u 1000 user
+RUN apt-get update \
+    && apt-get install -yq --no-install-recommends debootstrap \
+    && rm -rf /var/lib/apt/lists/* \
+    && debootstrap --variant minbase --include python3 eoan /chroot \
```
sroettger: is this all a single command so that the apt-get remove can clean up all files from the layer?

sirdarckcat: Yep, since we will only need the artifacts generated by the Dockerfile, this makes sure the image just contains what it needs.

base/nsjail-docker/Dockerfile
sirdarckcat: We dont. Unfortunately `upgrade` itself anyway gets cached, which means that the image always has updates as of the **first** time the last person to push built the image.

There are some hacks to do cache busting (eg, adding an argument FOO and running `apt-get upgrade -y && echo $FOO` at the end of every Dockerfile), so that you always get the image as of the last time the last person to push built it. This has the slight disadvantage that the image will grow over time with things it wouldn't need if it used the "upstream" images, but worse of all, apparently upgrade can sometimes fail inside of docker (we need to build the containers with --privileged it seems).

But generally, the suggested approach to trully keep things up-to-date is a Rapid-style worker that rebuilds and pushes skipping cache (with `--pull --no-cache`)  daily or more often. For a CTF, I suspect this will have to be a one time operation before the competition that happens at release time? Although I don't know.

Note that GCR has a vulnerability scanner, that checks for outdated packages. Maybe we can just surface it's findings as warnings when users deploy outdated packages.

sirdarckcat: Perhaps we can do a full build skipping cache when the challenges are marked as EXPOSED.

Maybe we can build the LoadBalancer (which anyway takes forever) earlier in the process, so that by the time we finish rebuilding, we hopefully don't have to wait so long.

sirdarckcat:
Thx! Pushing but feel free to revert if needed.

base/nsjail-docker/Dockerfile
sroettger: IMO upgrading once is still better than never at all.

We can have a LASTUPGRADE variable in the Dockerfile that is "never" by default.
Before the CTF starts, you just set this to a new value and check it into your git. All builds will then be at least that new.

sirdarckcat: take a look at the docker documentation. it explains that apt-get upgrade can break when not running with --privileged (which we don't).

`Avoid RUN apt-get upgrade and dist-upgrade, as many of the “essential” packages from the parent images cannot upgrade inside an unprivileged container. `

PR #40, opened 2020-03-09
Bypass the proof-of-work from the healthcheck

sroettger:
Use a per-CTF public/private key pair to sign pow requests.

Fixes #40

### DIFF
--- a/base/challenge-skeleton/challenge/config/nsjail.cfg
+++ b/base/challenge-skeleton/challenge/config/nsjail.cfg
@@ -38,6 +38,11 @@ mount: [
     src: "/config"
     dst: "/config"
     is_bind: true
+  },
+  {
+    src: "/pow-bypass"
+    dst: "/pow-bypass"
+    is_bind: true
   }
 ]
 
--- a/base/challenge-skeleton/healthcheck/exploit/doit.py
+++ b/base/challenge-skeleton/healthcheck/exploit/doit.py
@@ -19,7 +19,7 @@ import pwnlib
 def handle_pow(r):
     print(r.recvuntil('./pow.py solve '))
     challenge = r.recvline().strip()
-    p = pwnlib.tubes.process.process('bypass_pow.sh {}'.format(challenge))
+    p = pwnlib.tubes.process.process(['bypass_pow.sh', challenge])
     solution = p.readall().strip()
     r.sendline(solution)
     print(r.recvuntil('Correct\n'))
--- a/base/challenge-skeleton/healthcheck/exploit/doit.py
+++ b/base/challenge-skeleton/healthcheck/exploit/doit.py
@@ -14,6 +14,23 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from pwn import *
+import pwnlib
+
+def handle_pow(r):
+    r.recvuntil('./pow.py solve ')
+    challenge = r.recvline().strip()
+    proc = pwnlib.tubes.process.process('bypass_pow.sh')
+    solution = proc.readall().strip()
+    r.sendline(solution)
+    r.recvuntil('Correct\n')
+
+r = pwnlib.tubes.remote.remote('127.0.0.1', 1337)
+r.recvuntil('== proof-of-work: ')
+if r.recvline().startswith('enabled'):
+    handle_pow(r)
+
+r.sendline('cat /flag')
+r.recvuntil('CTF{')
+r.recvuntil('}')
 
 exit(0)
--- a/base/challenge-skeleton/healthcheck/exploit/doit.py
+++ b/base/challenge-skeleton/healthcheck/exploit/doit.py
@@ -17,20 +17,20 @@
 import pwnlib
 
 def handle_pow(r):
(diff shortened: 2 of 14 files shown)


base/nsjail-docker/Dockerfile
```diff
@@ -23,6 +23,12 @@ RUN /usr/sbin/useradd --no-create-home -u 1000 user
 COPY --from=bin /usr/bin/nsjail /usr/bin/nsjail
 COPY --from=chroot /chroot /chroot
 
+# create a python3 venv and install ecdsa for the proof-of-work
+RUN apt-get update && apt-get -y install python3-venv
+RUN python3 -m venv /venv
+RUN bash -c "source /venv/bin/activate && pip3 install ecdsa"
+RUN mv /venv /chroot/
+
```
sroettger: I had some trouble with nsjail and the module search path. I.e. nsjail changes the PATH variable and python3 doesn't find the modules installed in /usr/local anymore.
Instead of adding it to the variable, I figured it's cleaner to use a virtualenv instead.

sroettger:
fixes #40

base/nsjail-docker/Dockerfile
sirdarckcat: Shouldn't we put this in the base chroot? Since it's part of the core infrastructure.

base/challenge-skeleton/healthcheck/exploit/doit.py
```diff
@@ -14,6 +14,23 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from pwn import *
+import pwnlib
+
+def handle_pow(r):
```
sirdarckcat: Should this be a library? Like `from kctf_pow import handle_pow`

base/nsjail-docker/Dockerfile
sroettger: The base/nsjail-docker is part of the core infrastructure too no?

base/challenge-skeleton/healthcheck/exploit/doit.py
sroettger: I think it should stay as a subprocess since this is our API contract. And if someone wants to write a healthcheck in a different language, they will probably base it on this.

base/nsjail-docker/Dockerfile
sirdarckcat: I mean, cached. So we don't need users to build this, but we can do it for them.

base/nsjail-docker is modifiable per CTF, and needs to be manually upgraded with new releases.

base/challenge-skeleton/healthcheck/exploit/doit.py
```diff
@@ -14,6 +14,23 @@
...
+    r.recvuntil('./pow.py solve ')
+    challenge = r.recvline().strip()
```
sirdarckcat: Don't you need to send the challenge to proc?

base/challenge-skeleton/healthcheck/exploit/doit.py
```diff
@@ -14,6 +14,23 @@
...
+    proc = pwnlib.tubes.process.process('bypass_pow.sh')
```
sirdarckcat: Maybe instead of `proc` use a nicer name (bypass_pow?)

base/challenge-skeleton/healthcheck/exploit/doit.py
```diff
@@ -14,6 +14,23 @@
...
+    solution = proc.readall().strip()
+    r.sendline(solution)
+    r.recvuntil('Correct\n')
```
sirdarckcat: If this fails, will it be obvious by the error message what happened?

base/nsjail-docker/Dockerfile
sroettger: If you want me to create more cached images, then I have a feature request #44 :)

base/challenge-skeleton/healthcheck/exploit/doit.py
sroettger: oops, I think I tested this without enabling the proof of work ;)

base/nsjail-docker/Dockerfile
sirdarckcat: Haha ok :-)

base/challenge-skeleton/healthcheck/exploit/doit.py
sroettger: nope, but if this fails is probably an infrastructure problem

base/challenge-skeleton/challenge/config/nsjail.cfg
```diff
@@ -38,6 +38,11 @@ mount: [
     src: "/config"
     dst: "/config"
     is_bind: true
+  },
+  {
+    src: "/pow-bypass"
+    dst: "/pow-bypass"
+    is_bind: true
   }
```
sroettger: this is ugly, since we're exposing infrastructure details to the challenge directory.
An alternative for a future change could be to use an init container to copy the pow-bypass directory over or to have k8s mount to /chroot/pow-byass directly

PR #45, opened 2020-03-12
Updated google-cloud.md

goedi02:
I adapted this file to work well as a Tutorial in the Cloud Shell. I think there are a few details that could still be tweaked, e.g.:
- "Open pow.yaml and change the 0 to 1."
- "If it doesn't work, try again (or run the script)."

### DIFF
--- a/docs/google-cloud.md
+++ b/docs/google-cloud.md
@@ -1,11 +1,19 @@
 # CTF Infrastructure Walkthrough
 
-The purpose of this walkthrough is to guide you along the way to configure the infrastructure.
+Welcome to the CTF walkthrough!
 
-You can also open this walkthrough directly in [Google Cloud Shell](https://console.cloud.google.com/cloudshell/open?git_repo=https://github.com/google/kctf&tutorial=docs/google-cloud.md).
+## Goal of this walkthrough
 
-## First, enable the necessary GCP features
-Before we start, you need to have billing set up, and the compute API enabled. Let's do that now.
+The purpose of this walkthrough is to guide you through the configuration of the CTF infrastructure.
+
+Note: If not already doing so, you can also open this walkthrough directly in [Google Cloud Shell](https://console.cloud.google.com/cloudshell/open?git_repo=https://github.com/google/kctf&tutorial=docs/google-cloud.md).
+
+# Step 1 – Set up the cluster
+
+The first step consists of setting up the cluster and the related infrastructure.
+
+## Enable the necessary GCP features
+Set up billing, and enable the compute API:
 1. <walkthrough-project-billing-setup>Select a project and enable billing.</walkthrough-project-billing-setup>
 1. <walkthrough-enable-apis apis="compute.googleapis.com,container.googleapis.com,containerregistry.googleapis.com">Enable the compute API.</walkthrough-enable-apis>
 
@@ -14,7 +22,9 @@ You can enable APIs from the command line with:
 gcloud services enable compute container containerregistry.googleapis.com
 ```
 
-## Then, configure the project
+## Configure the project
+
+Perform the following steps to configure your project.
 
 ### Make sure your umask allows copying executable files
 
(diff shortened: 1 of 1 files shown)


docs/google-cloud.md
```diff
@@ -1,11 +1,19 @@
 # CTF Infrastructure Walkthrough
```
sirdarckcat: should we mention kCTF (the name of the project) here or google cloud? to clarify what this walkthrough is about. wdyt/?

docs/google-cloud.md
```diff
@@ -14,7 +22,9 @@ You can enable APIs from the command line with:
 gcloud services enable compute container containerregistry.googleapis.com
```
sirdarckcat: I think here we want to also include:
 - compute.googleapis.com
 - container.googleapis.com

docs/google-cloud.md
```diff
@@ -34,107 +44,121 @@ gcloud auth configure-docker
...
-This only needs to be done once per CTF. A "cluster" is essentially a group of VMs with a "master" that defines what to run there.
+It can take around a minute for the cluster to be created, during which the following message is displayed:
 
-This can take a while, you can expect to see `Creating cluster ... in ... Cluster is being health-checked...` and in a minute or so, you'll get a message telling you it's done.
+```Creating cluster ... in ... Cluster is being health-checked...``` 
+
+You'll get a message when cluster creation has completed.
 
 While you wait, here's how the infrastructure works:
-1. The CTF challenges will run inside "nsjail" (a security sandbox).
```
sirdarckcat: hmm, is this good way to present this? maybe we can instead send the user to the intro doc (the cluster takes so long, that doc is probably more appropriate?).

that said, the goal of this section is to take the opportunity of this "break" in the process to teach the reader something, but maybe we can skip this step if it's too long or doesnt make sense (we can let the cluster be created in the background). wdyt?

docs/google-cloud.md
```diff
@@ -34,107 +44,121 @@ gcloud auth configure-docker
...
-
-This all ensures the availability of the challenges, and saves some computing resources.
+1. If the VMs consume too little CPU, Kubernetes will shut down some VMs.
+1. Some very expensive challenges can be allocated to the same VMs as less busy challenges.
 
-Anyway, hopefully by now the cluster is already created and we can continue with the walkthrough.
+The above steps ensure the availability of the challenges, while using computing resources in a resourceful manner.
 
-If you are curious and have some more time, take a look at this [kCTF introduction](introduction.md), which includes a quick 8 minutes summary of what is kCTF and how it interacts with Kubernetes.
+By now, your cluster should be ready and we can continue with the walkthrough.
```
sirdarckcat: unfortunately, this is probably not gonna be true. this can take several minutes :(

sirdarckcat: see comment above for ideas

docs/google-cloud.md
```diff
@@ -34,107 +44,121 @@ gcloud auth configure-docker
...
+* Build a Docker image
+* Deploy the Docker image to the cluster
+* Expose the Docker image to the internet
 
-Click next to continue if your cluster is already created.
+Note: The cluster must be created before you continue, otherwise the following commands won't work. To create a cluster, see the preceding steps. Continue with the next steps if you already created a cluster.
 
-## First, call create_challenge.sh to copy the skeleton
-Run the following command to create a challenge called "demo-challenge"
+## Call create_challenge.sh to copy the skeleton
```
sirdarckcat: hmm seems like we changed the name of the command it's not called create_challenge.sh anymore

docs/google-cloud.md
```diff
@@ -34,107 +44,121 @@ gcloud auth configure-docker
...
 ```
 kctf-config-create
 ```
 
-### Configuration properties
-Type a path for storing the challenges
+### Set configuration properties
+Enter a path for storing the challenges:
 ```
 ~/demo-ctf-cluster
```
sirdarckcat: this needs to be "~/kctf-demo" for the command below to work

docs/google-cloud.md
```diff
@@ -34,107 +44,121 @@ gcloud auth configure-docker
...
-## First, call create_challenge.sh to copy the skeleton
-Run the following command to create a challenge called "demo-challenge"
+## Call create_challenge.sh to copy the skeleton
+Run the following command to create a challenge called "demo-challenge":
 ```
 kctf-chal-create demo-challenge
 ```
 
-This will create a directory called `demo-challenge` under the `kctf-demo` directory, and if you look inside of it (check files/chal for example), you'll find out the challenge configuration. The file in `files/chal` is the entry-point, that means it is executed every time there's a TCP connection. This demo challenge just runs bash, but a real challenge would instead expose a harder challenge that doesn't just give out a shell.
+This creates a directory called `demo-challenge` under the `kctf-demo` directory.
```
sirdarckcat: the directory is not kctf-demo atm, its demo-ctf-cluster. we need to change them so they work

docs/google-cloud.md
sroettger: kctf-chal-create?

docs/google-cloud.md
goedi02: Agree, this makes sense to add. I made some changes here at the beginning. Does it make sense?

docs/google-cloud.md
goedi02: Did I add it in the correct way?

docs/google-cloud.md
goedi02: sorry, the implication is not clear to me. Should I change the doc to "demo-ctf-cluster", or will you change the solution to use "kctf-demo"?

docs/google-cloud.md
goedi02: I agree this is kind of a weird "step", as the user does nothing but wait for an automatic process to complete. We could just add a sentence to the end of the "Configure the project" section saying "After configuring the project, the cluster is created automatically."

On the other hand: By having this separate step, we give the cluster creation more prominence. Maybe a way to decide is: if the info we currently provide in this section is really important for users to know, let's keep it. We could put it somewhere else and link to it, but I doubt many people will follow the link and moving to a new page is always harder than reading a bit of info integrated directly where you are. Thoughts?

goedi02: "we can skip this step if it's too long or doesnt make sense (we can let the cluster be created in the background)" what do you mean by this? This would be a change in the way the set up works?

docs/google-cloud.md
goedi02: we could rephrase: "Your cluster should soon be ready, when it is, you can continue with this walkthrough."

docs/google-cloud.md
goedi02: done

docs/google-cloud.md
goedi02: changed to kctf-chal-create

docs/google-cloud.md
sirdarckcat: I think so :)

docs/google-cloud.md
sirdarckcat: I think the info we provide here is important for the user to know, yes. Mainly because it's likely later on they'll need to know this to be able to fix things. Let's keep it then :)

docs/google-cloud.md
sirdarckcat: SGTM! lets do that :)

docs/google-cloud.md
sirdarckcat: I mean, the documentation was wrong before, and I just noticed =). We should make sure we use the same name for the kCTF walkthrough demo everywhere.

In other words, the step here (`~/kctf-demo`):
https://github.com/google/kctf/blob/5db06f7a02d93359753c4e50d964d78cd324b352/docs/google-cloud.md#set-configuration-properties

Should match the step here (`~/demo-ctf-cluster`):
https://github.com/google/kctf/blob/5db06f7a02d93359753c4e50d964d78cd324b352/docs/google-cloud.md#deploy-the-challenge

PR #46, opened 2020-03-16
strawman proposal for dir layout

sroettger:
Work in progress, creating pull request for discussion


base/challenge-skeleton/README.md
```diff
@@ -0,0 +1,41 @@
...
+* Makefile: anything that needs to be run before deploying a challenge
+* challenge: the files for the challenge container
+* healthcheck: the files for the healthcheck container (optional)
+* advanced: the kubernetes config used to deploy the challenge. E.g. add a tmpfs here or change the port.
+* .gen: a directory for files generated by make.
+
+## /challenge
+
+* image/: Dockerfile and files of the challenge image. Note that for defense in depth, you shouldn't put secrets in here.
+* dynamic: files to be added to the challenge at runtime. Secrets are ok, but note that there's a size limit. Supports live updates without a challenge restart.
```
sroettger: dynamic is a bad name. If I would read this, I wouldn't know what it's about without reading the README.

sroettger:
All of this could be so much simpler if we were ok with having secret files in the docker images.

base/challenge-skeleton/README.md
```diff
@@ -0,0 +1,41 @@
...
+* chal.conf: configure if the challenge should be deployed to the cluster if it's publicly accessible and if it has a healthcheck
+* Makefile: anything that needs to be run before deploying a challenge
+* challenge: the files for the challenge container
+* healthcheck: the files for the healthcheck container (optional)
+* advanced: the kubernetes config used to deploy the challenge. E.g. add a tmpfs here or change the port.
+* .gen: a directory for files generated by make.
+
+## /challenge
+
+* image/: Dockerfile and files of the challenge image. Note that for defense in depth, you shouldn't put secrets in here.
```
sroettger: the Dockerfile and contents should be in a separate directory so that we don't need a dockerignore. Otherwise we or users will make a mistake and secrets files can be sent to a build server.

sroettger: of course, unless we decide that we don't need to keep the docker images secret

base/challenge-skeleton/README.md
```diff
@@ -0,0 +1,41 @@
+# Quickstart guide to writing a challenge:
+
+* put the challenge binary in challenge/image/ and update the Dockerfile inside
+* by default /home/user/chal will be run. The command line can be changed in challenge/dynamic/nsjail.cfg.
+* Try the challenge locally with `make -j test-docker`
+* If you have a cluster ready, try it remotely with `make -j start && make port-forward`
```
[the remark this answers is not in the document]
sroettger: Hmm, let me time the difference and then decide

base/challenge-skeleton/README.md
```diff
@@ -0,0 +1,41 @@
...
+* challenge: the files for the challenge container
+* healthcheck: the files for the healthcheck container (optional)
+* advanced: the kubernetes config used to deploy the challenge. E.g. add a tmpfs here or change the port.
+* .gen: a directory for files generated by make.
+
+## /challenge
+
+* image/: Dockerfile and files of the challenge image. Note that for defense in depth, you shouldn't put secrets in here.
+* dynamic: files to be added to the challenge at runtime. Secrets are ok, but note that there's a size limit. Supports live updates without a challenge restart.
+* dynamic/kustomization.yaml: a kubernetes patch file. You only have to touch it if you add a new file to this directory.
```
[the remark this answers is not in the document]
sroettger: I think there was some reason why we had the kustomize.yaml, but I'm not sure what it was. Yeah, we should try to get rid of it

base/challenge-skeleton/README.md
sroettger: Ah, now I can reply again.
I don't like configuration because it wouldn't be clear why flag lives in config.
Maybe runtime-files or dynamic-files?
To me, "mount" would be intuitive since the stuff gets mounted into the container but I guess that depends on your background

sroettger: Changed it to runtime-files

base/challenge-skeleton/README.md
```diff
@@ -0,0 +1,41 @@
+# Quickstart guide to writing a challenge:
+
+* put the challenge binary in challenge/image/ and update the Dockerfile inside
+* by default /home/user/chal will be run. The command line can be changed in challenge/dynamic/nsjail.cfg.
```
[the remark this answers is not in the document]
sroettger: it's up to the user where to put that binary

base/challenge-skeleton/README.md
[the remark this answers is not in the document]
sroettger: removed -j since it also makes debugging harder, so probably not good advice in the quickstart

base/challenge-skeleton/README.md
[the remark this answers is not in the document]
sroettger: removing it in #48

sroettger:
Here's another proposal for a directory layout if we get rid of secrets completely and just embed them in the docker images.
The API contract would be:
* You need a challenge/image/Dockerfile that builds your chal container.
* You get port 1337 by default.
* The base image provides pow and nsjail, but it's up to you to use it. If you don't, you loose 1 layer of security and DoS protection and player isolation, so strongly recommended.
* You need to have a challenge/Makefile that provides a docker target. It needs to build the docker container and provide the id (and we just tag it).
* The healthcheck dir (optional) is exactly the same as challenge. Just that it needs to run a webserver on 8080 to handle /healthz requests. The base image provides pwntools and magic to bypass the pow.
* pow and chal.conf lives in config/. config/advanced has the k8s stuff.
* We should be able to guarantee few breaking changes if you don't touch config/advanced. And e.g. always make a new major version when there's a breaking change.

Here's how it would look like for the bash example:
```
challenge/
  image/
    chal
    Dockerfile
    flag
    nsjail.cfg
  src/ (optional)
    shell.c
  Makefile
config/
  advanced/
    deployment/
    network/
  chal.conf
  pow.conf
healthcheck/
  image/
    Dockerfile
    doit.py
    env
    healthz.py
    run.sh
  Makefile
Makefile -> ../kctf-conf/base/Makefile
```

base/challenge-skeleton/README.md
```diff
@@ -0,0 +1,40 @@
+# Quickstart guide to writing a challenge:
+
+* Put your challenge binary under challenge/image/ and update challenge/image/Dockerfile.
+* By default /chroot/home/user/chal will be run. The command line can be changed in challenge/runtime-files/nsjail.cfg.
```
sirdarckcat: We could make it so that the nsjail.cfg doesn't have the command to execute, and instead accept it from the command line. This way the user just has to do in the Dockerfile:
```
CMD drop_privs && nsjail /chroot/chal
```

base/challenge-skeleton/README.md
```diff
@@ -0,0 +1,40 @@
...
+* Try the challenge locally with `make test-docker`.
+* If you have a cluster ready, try it remotely with `make start && make port-forward`.
+
+## healthcheck
+
+* The healthcheck needs to run a webserver on port 8080 that responds to /healthz requests.
```
sirdarckcat: Will the skeleton provide an example web server?

Maybe we can make the port something else, like 911 or something to make it memorable as the healthcheck (I forgot 8080 was taken, and had issues with the Apache example)

base/challenge-skeleton/README.md
```diff
@@ -0,0 +1,40 @@
...
+ * 200 if the challenge is healthy, otherwise an error (e.g. 400).
```
sirdarckcat: Same as above, I accidentally deployed a web server in the challenge container in port 8080. Now I realize this also would make the healthcheck think it's working.

Maybe we should make it require responding with OK on /healthz.

base/challenge-skeleton/README.md
```diff
@@ -0,0 +1,40 @@
...
+* The base image comes preloaded with pwntools.
+* If your exploit is written in python, just drop it in healthcheck/exploit/doit.py.
+* Note: for defense in depth, the exploit is encrypted before being added to the docker image.
+
+# Directory layout:
+
+## /
+
+* chal.conf: Configure if the challenge should be deployed to the cluster if it's publicly accessible and if it has a healthcheck.
```
sirdarckcat: This is now in /config

base/challenge-skeleton/README.md
```diff
@@ -0,0 +1,40 @@
...
+* Makefile: Anything that needs to be run before deploying a challenge.
+* challenge: The files for the challenge container.
+* healthcheck: The files for the healthcheck container (optional).
+* advanced: The kubernetes config used to deploy the challenge. E.g. add a tmpfs here or change the port.
```
sirdarckcat: Now in /config

base/challenge-skeleton/README.md
```diff
@@ -0,0 +1,40 @@
...
+* .gen: A directory for files generated by make.
+
+## /challenge
+
+* image/: Dockerfile and files of the challenge image. Note that for defense in depth, you shouldn't put secrets in here.
+* runtime-files: Files to be added to the challenge at runtime. Secrets are ok, but note that there's a 1MB size limit for the whole directory. Supports live updates without a challenge restart.
```
sirdarckcat: This is now /config right?

base/challenge-skeleton/README.md
```diff
@@ -0,0 +1,40 @@
...
+* runtime-files/flag: The flag (duh).
+* runtime-files/nsjail.cfg: The configuration for the nsjail sandbox. Configure here limits and what you want to expose to the challenge chroot.
```
sirdarckcat: This can probably live in Dockerfile instead

base/challenge-skeleton/README.md
```diff
@@ -0,0 +1,40 @@
...
+* runtime-files/pow: Difficulty of the proof-of-work. 0 means disabled.
+
+## /healthcheck
+
+* /: The whole directory is optional. If you're not going to implement a healthcheck, feel free to delete this.
+* image/: Dockerfile and files of the healthcheck.
+* exploit/: The exploit code to check if the challenge is healthy. This directory will be encrypted for defense in depth.
```
sirdarckcat: We should explain where it will be mounted.

sirdarckcat:
I assume the flag would be in /config/flag.conf?

sroettger:
> I assume the flag would be in /config/flag.conf?
No, it would just be part of the docker image. So up to the user where to put it.

sirdarckcat:
there's still stuff missing from the strawman right?

base/challenge-skeleton/README.md
sroettger: re this and the other comments: all of this is outdated and was describing the "we need to use secrets" approach

sroettger:
what do you mean?

sirdarckcat:
was wondering if #51 implemented everyone from this proposal :)

sroettger:
closed in favor of #51

sroettger:
oh, just say you said not everything from here made it to #51, do you have an example?

PR #49, opened 2020-03-17
hide infrastructure files in chroot

sroettger:
Add a /chroot/.kctf directory and put all the files from the infrastructure there.

### DIFF
--- a/base/nsjail-docker/Dockerfile
+++ b/base/nsjail-docker/Dockerfile
@@ -30,9 +30,6 @@ RUN python3 -m venv /venv
 RUN bash -c "source /venv/bin/activate && pip3 install ecdsa"
 RUN mv /venv /chroot/
 
-#TODO should these exist by default? Or should the chal author specify mappings manually?
-RUN mkdir /chroot/config
-RUN mkdir /chroot/secrets
 RUN mkdir /chroot/.kctf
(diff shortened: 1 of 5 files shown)


base/nsjail-docker/Dockerfile
```diff
@@ -30,9 +30,12 @@ RUN python3 -m venv /venv
 RUN bash -c "source /venv/bin/activate && pip3 install ecdsa"
 RUN mv /venv /chroot/
 
+#TODO should these exist by default? Or should the chal author specify mappings manually?
```
sroettger: @sirdarckcat what do you think about this?
Should we stop mounting these directories by default and have the challenge author specify all mounts themselves?
Pro: no accidental leaks of files into the chroot
Pro: a cleaner chroot, everything is where it should be (flag at /flag for example)
Contra: extra step when adding a new file

sirdarckcat: I imagine it'll be confusing for most authors to make sense of whats gonna get mounted by k8s, so I think it's better if we do it for them as otherwise they won't use it

sroettger: ok, moved it to the challenge-skeleton dockerfile though since it's not needed by the base and users might want to clean it up.

PR #50, opened 2020-03-17
split infrastructure parts out of nsjail.cfg

sroettger:


base/challenge-skeleton/challenge/Dockerfile
```diff
@@ -21,4 +21,4 @@ RUN ln -s /secrets/flag /chroot/flag
 
 COPY files /chroot/home/user
 
-CMD kctf_setup && drop_privs nsjail --config /config/nsjail.cfg
+CMD kctf_setup && drop_privs nsjail --config /kctf/nsjail.cfg --config /config/nsjail.cfg
```
sirdarckcat: we probably should call them base.nsj and chal.nsj ? wdyt?

just something to clarify that one is the base and the other is on top.. since order matters

base/challenge-skeleton/challenge/config/nsjail.cfg
```diff
@@ -38,16 +38,6 @@ mount: [
     src: "/config"
     dst: "/config"
     is_bind: true
-  },
```
sirdarckcat: we dont need the other settings?

sirdarckcat: infra-wide i mean

base/challenge-skeleton/challenge/Dockerfile
sroettger: I don't think we should invent a new file extension so I would still go with .cfg as in the nsjail docs.
Maybe /kctf/nsjail_base.cfg and /config/nsjail.cfg?

base/challenge-skeleton/challenge/config/nsjail.cfg
sroettger: which other settings do you mean?

sirdarckcat: uidmap, chroot

base/challenge-skeleton/challenge/Dockerfile
sirdarckcat: Yea that works :-)

base/challenge-skeleton/challenge/config/nsjail.cfg
sroettger: ah, that doesn't work. See the thread with robert and wiktor

base/challenge-skeleton/challenge/Dockerfile
sroettger: done

PR #51, opened 2020-03-19
Simplify challenge directory layout

### LINKED ISSUE #12: reduce the number of files in the skeleton
- [ ] #13 replace makefiles to use `source ../Makefile`
- [ ] Remove the src directory
- [ ] Rename k8s to advanced
- [ ] Make config secret, move secrets to config

sroettger:
* challenge and healthcheck directories are simple Docker images now
* images now include the highest value secrets (exploits)
  * we need to make sure that they are properly secured
  * it should still require two bugs to reach them: 1) nsjail escape, 2) gain access to the node metadata server
  * that is, unless you have a kernel exploit but under that assumption it would have broken the previous version of the infrastructure as well

FIxes #12


sirdarckcat:
Are you planning more changes?

More generally, how should we manage updates to the samples? Also to documentation like SECURITY.md for ex

sroettger:
that should probably be part of this pull request. But before I do that work, I would like to discuss if these changes look good.

sroettger:
I will also add a README based on the one in the other pull request, wdyt?

sroettger:
@sirdarckcat updated the SECURITY.md and added a challenge/README.md, PTAL

SECURITY.md
```diff
@@ -97,26 +96,16 @@ As such, the risks on this edge are:
  - Project might be misconfigured
 
 Ways to limit these risks are:
- - kCTF limits the information in docker images to the minimum possible. For example, flags and exploits are not here.
- - kCTF could limit the access to the GCR images more tightly. Work to do this is tracked on [Issue 17](https://github.com/google/kctf/issues/17).
+ - kCTF could limits the information in docker images to the minimum possible. There are no plans to do this.
```
sirdarckcat: could limit (keine `s`)

base/challenge-skeleton/README.md
```diff
@@ -0,0 +1,52 @@
...
+* The base image comes preloaded with pwntools.
+* If your exploit is written in python, just drop it in healthcheck/image/doit.py.
+
+# Directory layout:
+
+## /config
+
+The config directory holds a few configuration files:
+
+* chal.conf: Configure if the challenge should be deployed to the cluster (for batch commands), if it's publicly accessible and if it has a healthcheck.
```
sirdarckcat: I would remove the (for batch commands), and then just make `make start` fail if this is not allowed in the configuration. I didn't know this was just for batch commands and feels a bit surprising.

docs/google-cloud.md
```diff
@@ -155,7 +155,7 @@ Now that you have a challenge up and running, you need to test it to make sure i
 Run the following command to connect to your challenge:
 
 ```
-telnet $(make ip) 1
```
sirdarckcat: netcat is not installed in cloudshell by default

sroettger: ok, reverted. Though it's not ideal since telnet has control characters so it will break for anything that is not ascii only

base/challenge-skeleton/README.md
sroettger: it already behaves this way, but I don't think that matters in practice. The main reason this option exists is for batch commands, at least I don't see another use case.

sroettger: removed

base/challenge-skeleton/README.md
```diff
@@ -0,0 +1,52 @@
...
+* A docker image will be built from challenge/image. For the simplest challenges, replacing challenge/image/chal will be enough.
+* Edit challenge/image/Dockerfile to change the commandline or which files should get included.
+* Try the challenge locally with `make test-docker`.
+* If you have a cluster ready, deploy the challenge with `make start`.
+  * To access it, create a port forward with `make port-forward` and connect via `nc localhost PORT` using the printed port.
+* Check out `make <tab>` for more commands.
+
+## healthcheck
+
+* The healthcheck needs to run a webserver on port 45281 that responds to /healthz requests.
```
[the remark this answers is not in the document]
sroettger: I figured if I choose a value myself, it will have the same issue as port 8080 that someone might reuse it, so I generated a random value.

base/challenge-skeleton/README.md
```diff
@@ -0,0 +1,52 @@
+# Quickstart guide to writing a challenge:
+
+* A docker image will be built from challenge/image. For the simplest challenges, replacing challenge/image/chal will be enough.
```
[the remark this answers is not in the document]
sroettger: done

SECURITY.md
sroettger: done

sroettger:
I would submit this as is and then update the samples in follow up commits, wdyt?

sirdarckcat:
Maybe samples can be updated on every release? Rather than every commit.

sroettger:
every change that requires an update to the samples would be a new release anyway if there are breaking changes

PR #52, opened 2020-03-19
create a namespace per challenge

sroettger:
This does:
* one namespace per challenge
This does *not*:
* allow multiple challenges to share a namespace

Some thoughts on how we could support the second part:
* leave in all the complicated logic to change the names of everything
* generate chal.conf via kctf-chal-create and put the namespace name in
* use the namespace from chal.conf
Doing these things adds more complexity of course, so we should decide if we want this.

Related to #8

### DIFF
--- a/base/Makefile
+++ b/base/Makefile
@@ -58,15 +58,12 @@ else
 endif
 
 stop: .cluster-config
-	for resource_type in deployment service hpa; do
-	  kubectl get "$${resource_type}/${CHALLENGE_NAME}" >/dev/null 2>&1 && kubectl delete "$${resource_type}/${CHALLENGE_NAME}" || true
-	done
-	kubectl get "configMap/${CHALLENGE_NAME}-pow" >/dev/null 2>&1 && kubectl delete "configMap/${CHALLENGE_NAME}-pow" || true
+	kubectl delete namespace ${CHALLENGE_NAME}
 
 ip: .cluster-config
 	LB_IP=""
 	while [ -z "$${LB_IP}" ]; do
-	  LB_IP=$$(kubectl get "service/${CHALLENGE_NAME}" -o=jsonpath='{.status.loadBalancer.ingress[0].ip}')
+	  LB_IP=$$(kubectl get "service/chal" -o=jsonpath='{.status.loadBalancer.ingress[0].ip}')
 	  sleep 3
 	done
 	echo "$${LB_IP}"
@@ -102,13 +99,13 @@ healthcheck-logs: .cluster-config
 	kubectl logs -l "app=${CHALLENGE_NAME}" -c healthcheck
 
 ssh: .cluster-config
-	kubectl exec deployment/${CHALLENGE_NAME} -c challenge -it /bin/bash
+	kubectl exec deployment/chal -c challenge -it /bin/bash
 
 healthcheck-ssh: .cluster-config
-	kubectl exec deployment/${CHALLENGE_NAME} -c healthcheck -it /bin/bash
+	kubectl exec deployment/chal -c healthcheck -it /bin/bash
 
 port-forward: .cluster-config
-	kubectl port-forward deployment/${CHALLENGE_NAME} :1337 &
+	kubectl port-forward deployment/chal :1337 &
 
 test-docker: .gen/docker-id
 	docker ps -f "id=$$(cat .gen/docker-id)"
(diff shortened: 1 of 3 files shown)


base/Makefile
```diff
@@ -249,50 +248,16 @@ ${CLUSTER_GEN}/remote-healthcheck-image: ${CLUSTER_GEN}/healthcheck-image-tagged
...
+	kubectl config set-context --current --namespace="${CHALLENGE_NAME}"
+	kubectl create namespace "${CHALLENGE_NAME}" --dry-run -oyaml | kubectl apply -f -
+	kubectl patch ServiceAccount default --patch "automountServiceAccountToken: false"
 	mkdir -p ${CLUSTER_GEN} >&2
 
 .FORCE:
 
 define DEPLOYMENT_KUSTOMIZATION
 commonLabels:
   app: "$(CHALLENGE_NAME)"
```
sroettger: @sirdarckcat this can be removed as well, right?

base/Makefile
```diff
@@ -57,15 +57,12 @@ else
 endif
 
 stop: .cluster-config
-	for resource_type in deployment service hpa; do
-	  kubectl get "$${resource_type}/${CHALLENGE_NAME}" >/dev/null 2>&1 && kubectl delete "$${resource_type}/${CHALLENGE_NAME}" || true
-	done
-	kubectl get "configMap/${CHALLENGE_NAME}-pow" >/dev/null 2>&1 && kubectl delete "configMap/${CHALLENGE_NAME}-pow" || true
+	kubectl delete namespace ${CHALLENGE_NAME}
```
sroettger: challenge stop is really nice now :)

base/Makefile
```diff
@@ -249,50 +248,16 @@ ${CLUSTER_GEN}/remote-healthcheck-image: ${CLUSTER_GEN}/healthcheck-image-tagged
 	@  exit 1
 	@fi
 	kubectl config use-context "kctf_${PROJECT}_${ZONE}_${CLUSTER_NAME}" >&2
+	kubectl config set-context --current --namespace="${CHALLENGE_NAME}"
```
sroettger: I'm setting the namespace during config load. This has the nice side effect that kctf-kubectl still works without specifying a namespace since it defaults to the last challenge that you touched.

base/Makefile
sirdarckcat: Which part? Not sure I understood :-)

base/Makefile
sirdarckcat: Do we still delete the load balancer?

base/Makefile
sirdarckcat: We probably need to update the troubleshooting playbook then. How do you switch to another task?

sirdarckcat:
A MySQL + Apache challenge would have a single Pod with two containers, or two Pods? If we want to support independent scaling of MySQL I think we need to support somehow more than one Pod per challenge, **but** that has significant isolation consequences.

I think for now it's safer to ask authors to bootstrap all challenge setup inside a single Pod, with multiple containers.

base/Makefile
sroettger: I think I made a mistake highlighting the line. All the code that sets/uses the app=challengename is not needed since app == namespace now, right?

base/Makefile
sroettger: I think so, that one should be namespaced too, right?

base/Makefile
sroettger: either allways set --namespace=challengename, run the config set-context command manually, or run any make command in the challenge dir, e.g. make status.
What would you recommend?

sroettger:
Should we support multiple containers out of the box or would you have the players mess with the k8s config in config/advanced for these cases?

base/Makefile
sirdarckcat: ah yea, I think so =D

base/Makefile
sirdarckcat: okeey :)

base/Makefile
sirdarckcat: --namespace=challengename is probably ok :) similar to before

sirdarckcat:
> Should we support multiple containers out of the box or would you have the players mess with the k8s config in config/advanced for these cases?

Maybe it can be as easy as just copying the challenge directory? We just build/tag all directories.

base/Makefile
sirdarckcat: yea, sadly.. this part of restarting isn't super nice. I think we should just avoid having people do `make stop` then, and treat every case where that's needed as a bug =(

sroettger:
the k8s part will be a bit more complicated since for every new container, we would have to generate a patch for the deployment to add it. But would definitely be possible if the need arises.
Anway, going to merge this once the other pull request is in.

sirdarckcat:
makes sense, yea.

base/Makefile
sroettger: the app thing can't be removed since the deployment needs a selector and that one usually uses the app label

PR #54, opened 2020-03-23
Small copy changes

sirdarckcat:
@goedi02 feel free to make changes directly to the copy-readme branch here:
https://github.com/google/kctf/edit/copy-readme/base/challenge-skeleton/README.md

### DIFF
--- a/base/challenge-skeleton/README.md
+++ b/base/challenge-skeleton/README.md
@@ -9,16 +9,6 @@ The basic steps when preparing a challenge are:
   * To access the challenge, create a port forward with `make port-forward` and connect to it via `nc localhost PORT` using the printed port.
 * Check out `make <tab>` for more commands.
 
-## Healthcheck
-
-Important details regarding healthchecks:
-
-* Run the healthcheck on a webserver on port 45281 that responds to `/healthz` requests.
- * The healthcheck returns 200 if the challenge is healthy, otherwise an error (e.g. 400).
- * You can find an example webserver in `healthcheck/image/healthz.py`.
-* The base image comes preloaded with pwntools.
-* If your exploit is written in Python, drop it in `healthcheck/image/doit.py`.
-
 ## Directory layout
 
 The following directories are available:
@@ -27,7 +17,7 @@ The following directories are available:
 
 The `config` directory holds a few configuration files:
 
-* `chal.conf`: Configure this file if you are deploying the challenge to the cluster, if it's publicly accessible, and if it has a healthcheck.
+* `chal.conf`: Configure this file if you are deploying the challenge to the cluster, if it's publicly accessible, and/or if it has a healthcheck.
 * `pow.conf`: Contains the difficulty of the proof-of-work, 0 means disabled.
 * `advanced`: The Kubernetes config used to deploy the challenge. E.g. you can add a tmpfs here or change the port.
 
@@ -43,15 +33,27 @@ The `challenge` directory contains the challenge binary and anything else relate
 The `healthcheck` directory is optional. If you don't want to write a healthcheck, feel free to delete the directory. However, we strongly recommend that you implement a healthcheck :).
 
 * `image/`: Dockerfile and files of the healthcheck.
-* `Makefile`: Edit this file if you need to run code before running your healthcheck.
+* `Makefile`: Edit this file if you need to build any custom files before running your healthcheck.
+
+
+#### Healthcheck
+
+Important details regarding healthchecks:
+
+* Run the healthcheck on a webserver on port 45281 that responds to `/healthz` requests.
+ * The healthcheck returns 200 if the challenge is healthy, otherwise an error (e.g. 400).
+ * You can find an example webserver in `healthcheck/image/healthz.py`.
+* The base image comes preloaded with pwntools.
+* If your exploit is written in Python, drop it in `healthcheck/image/doit.py`.
 
 ## API contract
 
-Ensure the directories described above fulfill the following requirements to ensure they work with kCTF:
+Ensure your setup fulfills the following requirements to ensure it works with kCTF:
 
+* In your cmdline, call `kctf_setup` as the first command.
 * You can do pretty much whatever you want in the `challenge` directory but:
   * You need to have a `Makefile` with the `.gen/docker-image` target that builds a Docker image.
-  * We strongly recommend to use nsjail in all challenges.
+  * We strongly recommend using nsjail in all challenges.
 * Your challenge receives connections on port 1337.
 * The healthcheck directory is optional.
   * If it exists, the image should run a webserver on port 45281 and respond to `/healthz` requests.
--- a/base/challenge-skeleton/README.md
+++ b/base/challenge-skeleton/README.md
@@ -5,7 +5,7 @@ The basic steps when preparing a challenge are:
 * A Docker image is built from `challenge/image`. For the simplest challenges, replacing `challenge/image/chal` is sufficient.
 * Edit `challenge/image/Dockerfile` to change the commandline or the files you want to include.
 * Try the challenge locally with `make test-docker`.
-* If you have prepared a cluster, deploy the challenge with `make start`.
+* If you have already prepared a [cluster](https://github.com/google/kctf/blob/master/docs/google-cloud.md), deploy the challenge with `make start`.
   * To access the challenge, create a port forward with `make port-forward` and connect to it via `nc localhost PORT` using the printed port.
 * Check out `make <tab>` for more commands.
 
@@ -26,14 +26,14 @@ The `config` directory holds a few configuration files:
 The `challenge` directory contains the challenge binary and anything else related to the challenge:
 
 * `image/`: Dockerfile and files of the challenge image.
-* `Makefile`: Edit this file if you need to run code before building the Docker image, e.g. for building the challenge from source.
+* `Makefile`: Edit this file if you need to run code when building the Docker image, e.g. for building the challenge from source.
 
 ### /healthcheck
 
 The `healthcheck` directory is optional. If you don't want to write a healthcheck, feel free to delete the directory. However, we strongly recommend that you implement a healthcheck :).
 
 * `image/`: Dockerfile and files of the healthcheck.
-* `Makefile`: Edit this file if you need to build any custom files before running your healthcheck.
+* `Makefile`: Edit this file if you need to build any custom files when building your healthcheck.
 
 
 #### Healthcheck
@@ -50,10 +50,10 @@ Important details regarding healthchecks:
 
 Ensure your setup fulfills the following requirements to ensure it works with kCTF:
 
-* In your cmdline, call `kctf_setup` as the first command.
+* Verify `kctf_setup` is used as the first command in the CMD instruction of your [Dockerfile](https://github.com/google/kctf/blob/master/base/challenge-skeleton/challenge/image/Dockerfile).
 * You can do pretty much whatever you want in the `challenge` directory but:
-  * You need to have a `Makefile` with the `.gen/docker-image` target that builds a Docker image.
-  * We strongly recommend using nsjail in all challenges.
+  * In the [Makefile](https://github.com/google/kctf/blob/master/base/challenge-skeleton/challenge/Makefile), the `.gen/docker-image` target that builds a Docker image must be configured. The configuration is set by default and usually does not require changes.
+  * We strongly recommend using nsjail in all challenges. While NsJail is already installed, you need to configure it [here](https://github.com/google/kctf/blob/master/base/challenge-skeleton/challenge/image/nsjail.cfg). For more information on nsjail, see the [official website](https://nsjail.dev/).
 * Your challenge receives connections on port 1337.
 * The healthcheck directory is optional.
   * If it exists, the image should run a webserver on port 45281 and respond to `/healthz` requests.
--- a/base/challenge-skeleton/README.md
+++ b/base/challenge-skeleton/README.md
@@ -1,52 +1,59 @@
-# Quickstart guide to writing a challenge:
+# Quickstart guide to writing a challenge
 
-* A docker image will be built from `challenge/image`. For the simplest challenges, replacing `challenge/image/chal` will be enough.
-* Edit `challenge/image/Dockerfile` to change the commandline or which files should get included.
+The basic steps when preparing a challenge are:
+
+* A Docker image is built from `challenge/image`. For the simplest challenges, replacing `challenge/image/chal` is sufficient.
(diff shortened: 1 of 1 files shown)


sirdarckcat:
posting the questions from dirk to the thread:


> Some questions to the content:
> - chal.conf - needs to be configured only if all three of the mentioned criteria are fulfilled? or should it be "or"?
> - is the description I added for "Makefile" in the "/healthcheck" section correct?
> - API contract - is the reference to the three directories in the intro sentence ok?
> - Generally, the API contract section contains infos that refer to the directories, would it make sense to move that information to those sections directly? That would leave just "Your challenge receives connections on port 1337." for this section. 
> - Should we nest the "Healthcheck" section below the section on the "/healthcheck" directory?

sirdarckcat:
> posting the questions from dirk to the thread:
> 
> > Some questions to the content:
> > 
> > * chal.conf - needs to be configured only if all three of the mentioned criteria are fulfilled? or should it be "or"?
it should always be there.

> > * is the description I added for "Makefile" in the "/healthcheck" section correct?
I think it should be "edit this file if you need to build any custom files, or something like that. @sroettger wdyt?

> > * API contract - is the reference to the three directories in the intro sentence ok?
lgtm

> > * Generally, the API contract section contains infos that refer to the directories, would it make sense to move that information to those sections directly? That would leave just "Your challenge receives connections on port 1337." for this section.
I think @sroettger wanted to delineate the difference between what can break between versions, and what wont. That said, maybe it's not clear by "API contract" alone.

> > * Should we nest the "Healthcheck" section below the section on the "/healthcheck" directory?
that makes sense, @sroettger wdyt?

sroettger:
> > posting the questions from dirk to the thread:
> > > Some questions to the content:
> > > 
> > > * chal.conf - needs to be configured only if all three of the mentioned criteria are fulfilled? or should it be "or"?
> 
> it should always be there.
> 
> > > * is the description I added for "Makefile" in the "/healthcheck" section correct?
> 
> I think it should be "edit this file if you need to build any custom files, or something like that. @sroettger wdyt?

Yeah sounds good.

> 
> > > * API contract - is the reference to the three directories in the intro sentence ok?
> 
> lgtm
> 
> > > * Generally, the API contract section contains infos that refer to the directories, would it make sense to move that information to those sections directly? That would leave just "Your challenge receives connections on port 1337." for this section.
> 
> I think @sroettger wanted to delineate the difference between what can break between versions, and what wont. That said, maybe it's not clear by "API contract" alone.
> 
> > > * Should we nest the "Healthcheck" section below the section on the "/healthcheck" directory?
> 
> that makes sense, @sroettger wdyt?

SGTM!

base/challenge-skeleton/README.md
```diff
@@ -1,52 +1,59 @@
...
-The healthcheck directory is optional. If you don't want to write a healthcheck, feel free to delete it. You really should have healthchecks though :).
+The `healthcheck` directory is optional. If you don't want to write a healthcheck, feel free to delete the directory. However, we strongly recommend that you implement a healthcheck :).
 
-* image/: Dockerfile and files of the healthcheck.
-* Makefile: same as for the challenge directory.
+* `image/`: Dockerfile and files of the healthcheck.
+* `Makefile`: Edit this file if you need to run code before running your healthcheck.
 
-# API contract
+## API contract
```
sroettger: I forgot one point: In your cmdline, you have to call `kctf_setup` as the first command. Would you mind adding that point as part of this pull request?

sroettger: thank you! @goedi02

base/challenge-skeleton/README.md
```diff
@@ -1,52 +1,61 @@
...
-* A docker image will be built from `challenge/image`. For the simplest challenges, replacing `challenge/image/chal` will be enough.
-* Edit `challenge/image/Dockerfile` to change the commandline or which files should get included.
+The basic steps when preparing a challenge are:
+
+* A Docker image is built from `challenge/image`. For the simplest challenges, replacing `challenge/image/chal` is sufficient.
+* Edit `challenge/image/Dockerfile` to change the commandline or the files you want to include.
 * Try the challenge locally with `make test-docker`.
-* If you have a cluster ready, deploy the challenge with `make start`.
-  * To access it, create a port forward with `make port-forward` and connect via `nc localhost PORT` using the printed port.
+* If you have prepared a cluster, deploy the challenge with `make start`.
```
sirdarckcat: Is it clear what it means to have prepared a cluster? Maybe we can say instead something like "if you have prepared a [cluster](link to how to setup a cluster) already, ..." Or maybe we can just ignore this bullet.

base/challenge-skeleton/README.md
```diff
@@ -1,52 +1,61 @@
...
+  * To access the challenge, create a port forward with `make port-forward` and connect to it via `nc localhost PORT` using the printed port.
```
sirdarckcat: Port forwarding only works inside a cluster, so maybe we should just use instructions here that work for local dev?

base/challenge-skeleton/README.md
```diff
@@ -1,52 +1,61 @@
...
+### /challenge
 
-* chal.conf: Configure if the challenge should be deployed to the cluster, if it's publicly accessible and if it has a healthcheck.
-* pow.conf: Contains the difficulty of the proof-of-work, 0 means disabled.
-* advanced: The kubernetes config used to deploy the challenge. E.g. you can add a tmpfs here or change the port.
+The `challenge` directory contains the challenge binary and anything else related to the challenge:
 
-## /challenge
+* `image/`: Dockerfile and files of the challenge image.
+* `Makefile`: Edit this file if you need to run code before building the Docker image, e.g. for building the challenge from source.
```
sirdarckcat: Instead of "before building" maybe it should when "when building", as Makefile is the one that technically builds the challenge.

base/challenge-skeleton/README.md
```diff
@@ -1,52 +1,61 @@
...
-The challenge binary and anything else challenge related goes in here.
+### /healthcheck
 
-* image/: Dockerfile and files of the challenge image.
-* Makefile: You can edit this file if you need to run code before building the docker image, e.g. for building the challenge from source.
+The `healthcheck` directory is optional. If you don't want to write a healthcheck, feel free to delete the directory. However, we strongly recommend that you implement a healthcheck :).
 
-## /healthcheck
+* `image/`: Dockerfile and files of the healthcheck.
+* `Makefile`: Edit this file if you need to build any custom files before running your healthcheck.
```
sirdarckcat: Same as before, maybe instead of "before running" it should say "when building".

base/challenge-skeleton/README.md
```diff
@@ -1,52 +1,61 @@
...
+## API contract
 
-Here are the requirements what this directory must look like to work with kCTF:
+Ensure your setup fulfills the following requirements to ensure it works with kCTF:
 
-* You can do pretty much whatever you want in the challenge dir but:
-  * You need to have a Makefile with the .gen/docker-image target that builds a docker image.
-  * We strongly recommend to use nsjail in all challenges.
-* Your challenge will receive connections on port 1337.
+* In your cmdline, call `kctf_setup` as the first command.
```
sirdarckcat: Maybe we should add an example here. It's not clear to me what this means.
```Dockerfile
CMD kctf_setup && kctf_nsjail /home/user/chal
```

base/challenge-skeleton/README.md
```diff
@@ -1,52 +1,61 @@
...
+* You can do pretty much whatever you want in the `challenge` directory but:
+  * You need to have a `Makefile` with the `.gen/docker-image` target that builds a Docker image.
```
sirdarckcat: Same as before, can we add an example? And explain what that file should contain
```Makefile
.gen/docker-image:
docker build . >$@
```

base/challenge-skeleton/README.md
```diff
@@ -1,52 +1,61 @@
...
+  * We strongly recommend using nsjail in all challenges.
```
sirdarckcat: Should we explain here how? @sroettger

base/challenge-skeleton/README.md
sroettger: adding a link sounds good

base/challenge-skeleton/README.md
sroettger: that's why it's indented, you think that's not clear?

base/challenge-skeleton/README.md
sroettger: makes sense

base/challenge-skeleton/README.md
sroettger: or link to the Dockerfile?
We could also say "leave kctf_setup as the first command in the Dockerfile" or similar

base/challenge-skeleton/README.md
sroettger: wouldn't it be better to link to the Makefile instead?

base/challenge-skeleton/README.md
sroettger: How would you explain it? We could give the example again, but that's already in the Dockerfile of the skeleton

base/challenge-skeleton/README.md
sirdarckcat: Cool, maybe we can link to the google-cloud.md walkthrough

base/challenge-skeleton/README.md
sirdarckcat: Ok, I missed the indentation :-)

base/challenge-skeleton/README.md
sirdarckcat: Yea let's do that :-)

base/challenge-skeleton/README.md
sirdarckcat: sg

base/challenge-skeleton/README.md
sirdarckcat: Maybe we can say that "NsJail will be installed already, and you can configure it here" and rather than recommending people to use NsJail, just assume they will and explain how to use it.

base/challenge-skeleton/README.md
sirdarckcat: https://github.com/google/kctf/blob/master/base/challenge-skeleton/Makefile

base/challenge-skeleton/README.md
sirdarckcat: @goedi02 wdyt?

base/challenge-skeleton/README.md
sirdarckcat: @goedi02 do you think it's clear? Not everyone might be familiar with Makefiles. If we linked you to that file would you know what we mean by this sentence?

base/challenge-skeleton/README.md
sirdarckcat: @goedi02 unlike the Makefile below, I think the Dockerfile would be clear (it's less magical than the Makefile below as it's not customized too much).

goedi02: ok, so we could write: "Ensure kctf_setup is used as the first command in your [Dockerfile](https://github.com/google/kctf/blob/master/base/challenge-skeleton/challenge/image/Dockerfile)"

base/challenge-skeleton/README.md
goedi02: Agree, if NsJail requires configuration, we need to emphasize that fact. So how about:

"We strongly recommend using NsJail in all challenges. While NsJail is already installed, you need to configure it [here](link)."

The above suggestion assumes that the question "how to configure" is obvious from the file we are linking to. If not, we could add details to the above sentence:

"...you need to configure it [here](link) by adding/specifying XZY."

base/challenge-skeleton/README.md
goedi02: @sirdarckcat not a 100% clear to me. In the file we link to, I can see ".gen/docker-image" is already there. But I assume I need to customize/modify that for my use case? We ought to emphasize the action:

"In the [Makefile](https://github.com/google/kctf/blob/master/base/challenge-skeleton/challenge/Makefile): Ensure the `.gen/docker-image` target that builds a Docker image is configured according to your setup."

- can we assume people know what they need to change when configuring? Or should we add more detail on that?

base/challenge-skeleton/README.md
sirdarckcat: We would be linking to an example configuration, and if people want to go more advanced they would have to visit nsjail's website https://nsjail.dev/

btw, I think we should call it nsjail without the camelcase :P I dunno why I used NsJail before, but the site uses it in all lowercase.

base/challenge-skeleton/README.md
sirdarckcat: I think most users won't need to change it, so maybe we can mention that by default the makefile is configured to build a docker image?

base/challenge-skeleton/README.md
sirdarckcat: maybe people might get confused as a Dockerfile has many things (RUN, FROM, COPY, etc) and here we are talking about the "command" (CMD). Maybe we can explicitly call out "Ensure kctf_setup is used as the first command in your Dockerfile's CMD instruction?" or something like that.
https://docs.docker.com/engine/reference/builder/#cmd has some documentation about what CMD is

base/challenge-skeleton/README.md
goedi02: Done

base/challenge-skeleton/README.md
goedi02: Done

base/challenge-skeleton/README.md
goedi02: Added link

base/challenge-skeleton/README.md
goedi02: Thank you for the clarification. We definitely need to be explicit about the CMD instruction, agreed! Changed sentence, including link to the Dockerfile.

base/challenge-skeleton/README.md
goedi02: Ok, rephrased this to: "In the Makefile, the .gen/docker-image target that builds a Docker image must be configured. The configuration is set by default and usually does not require changes."

sirdarckcat:
@goedi02 moved you to assignee from reviewers :)

base/challenge-skeleton/README.md
goedi02: done. The new text contains a link to the [nsjail.cfg](https://github.com/google/kctf/blob/master/base/challenge-skeleton/challenge/image/nsjail.cfg) as well as to the [nsjail website](https://nsjail.dev/).

goedi02:
@sirdarckcat if you are good with the current state, let's "squash and merge". Let me know :-)

PR #55, opened 2020-03-24
Update xss-bot to use new directory layout

sroettger:


sroettger:
Ah, I just saw I should probably update all of them together since they share the kctf-conf directory

base/nsjail-docker/files/drop_privs
```diff
@@ -2,4 +2,4 @@
 
 # There are two copies of this file in the nsjail and healthcheck base images.
 
-exec setpriv --init-groups --reset-env --reuid user --regid user --inh-caps=-all -- $@
+exec setpriv --init-groups --reset-env --reuid user --regid user --inh-caps=-all -- "$@"
```
sirdarckcat: Why do you need to quote it? Doesn't that make command line arguments the name of the command?

samples/xss-bot/healthcheck/Makefile
```diff
@@ -0,0 +1,26 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+SHELL := bash
+
+# IMPORTANT
+# - This will be the entrypoint when building the docker image.
+.gen/docker-image: .gen/new-image-id
+	NEW_ID=$$(cat .gen/new-image-id | cut -d ':' -f 2); if [[ "$${NEW_ID}" != $$(cat $@ 2>/dev/null) ]]; then echo "$${NEW_ID}" > $@; fi
```
sirdarckcat: People actually need to type all this for every task? :-/

sroettger: I'm not sure if I understand, when would they type this?

base/nsjail-docker/files/drop_privs
sroettger: "$@" will be replaced with "$1" "$2" ...
It's magic :)

samples/xss-bot/healthcheck/Makefile
sirdarckcat: when making a new task, I mean. it looks a bit too magical.

can we instead just use the command below as the only required rule?

base/nsjail-docker/files/drop_privs
sirdarckcat: cool!

samples/xss-bot/healthcheck/Makefile
sroettger: when making a new task, they write kctf-chal-create $chal_name. That copies the Makefile from the challenge skeleton in here so there's no typing involved.
Do you mean it's an ugly command to have in here and if we could move it to the global makefile?

sirdarckcat: yea, so having something like:
```
.gen/new-image-id:
docker build . > $@
```
is clear what is doing and why, but the bash comparison/cutting and such is not super trivial to understand

sroettger: yeah, good idea.

sroettger:
continuing in #60

samples/xss-bot/healthcheck/Makefile
sroettger: done (see #60)

PR #67, opened 2020-04-16
Update troubleshooting.md

goedi02:
Question for reviewers: Should I move the information from the "conclusion" to the start of the file? To me this seems to be information that would be useful to know from the beginning. It could be integrated into the list of the three environments in the intro.

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -3,9 +3,9 @@
 Users are reporting a challenge is down, the author is offline, and you don't know how long the challenge has been broken. The challenge had no healthcheck configured, and there's no documentation. Someone forgot to test the task. A nightmare come true.
 
 This guide will show you how to troubleshoot a broken challenge, assuming you don't know how the challenge works. The guide is divided into three parts that take you through troubleshooting a task in different environments:
-* [In Docker](#troubleshooting-with-docker)
-* [In a local (Kubernetes) cluster](#troubleshooting-with-kubernetes)
-* [Remotely](#troubleshooting-remotely)
+* [In Docker](#troubleshooting-with-docker) – Testing locally with `make test-docker` is the easiest and fastest way to troubleshoot challenges, and you don't need to know much apart from some basic Docker commands presented in the following section.
+* [In a local (Kubernetes) cluster](#troubleshooting-with-kubernetes) – Troubleshooting with Kubernetes requires more setup, but is needed only on rare occasions, as it is only relevant if the author made changes to the Kubernetes setup, or if there's a bug in kCTF.
+* [Remotely](#troubleshooting-remotely) – Troubleshooting remotely is trivial, although it runs the risk of the user leaving the remote state in an inconsistent state, however it's a good last-resort.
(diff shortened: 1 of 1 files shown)


docs/troubleshooting.md
```diff
@@ -94,11 +95,11 @@ This should trigger the same error as the one found in `docker logs` from the pr
 /usr/bin/nsjail --config /config/nsjail.cfg -Mo -v
 ```
 
-This should output more errors, that should provide you with more details about what went wrong.
+This should output more errors, and these errors should provide more details about what went wrong.
 
 ## Troubleshooting with Kubernetes
 
-If everything works in docker, the problem might be higher up (in Kubernetes). The first step to debug this would be to check if the challenge works in a local cluster. Follow the instructions [here](local-testing.md#running-the-challenge-in-kubernetes) for getting the task running in KIND.
+If everything works in Docker, the problem might be higher up (in Kubernetes). The first step to debug this would be to check if the challenge works in a local cluster. Follow the instructions [here](local-testing.md#running-the-challenge-in-kubernetes) for getting the task running in KIND.
```
sroettger, line 102: for KIND, maybe link to https://github.com/kubernetes-sigs/kind

docs/troubleshooting.md
```diff
@@ -60,18 +61,18 @@ If the challenge is a pwnable, you might see an error formatted like this:
 [E][2020-02-02T20:20:02+0200][1] void subproc::subprocNewProc(nsjconf_t*, int, int, int, int)():204 execve('/bin/sh') failed: No such file or directory
 ```
 
-That's an NsJail error. In this case, it is complaining that `/bin/sh` failed and gave the error *"No such file or directory"*.
+That's an nsjail error. In this case, it is complaining that `/bin/sh` failed and returned the error *No such file or directory*.
```
sroettger, line 64: I also think nsjail looks better in lowercase (and the command is written this way) but just fyi https://github.com/google/nsjail uses the NsJail capitalization.
But still, +1 for lower case :)

sroettger:
> Question for reviewers: Should I move the information from the "conclusion" to the start of the file? To me this seems to be information that would be useful to know from the beginning. It could be integrated into the list of the three environments in the intro.

Yeah, I think that would be useful in the beginning.

docs/troubleshooting.md
goedi02: Done

docs/troubleshooting.md
goedi02: Thanks for the info! I agree with sticking with lower case here; as long as the kCTF  doc is consistent, it's fine. When reviewing the Google Cloud walkthrough with Eduardo, we also opted for the lower case spelling.

goedi02:
> > Question for reviewers: Should I move the information from the "conclusion" to the start of the file? To me this seems to be information that would be useful to know from the beginning. It could be integrated into the list of the three environments in the intro.
> 
> Yeah, I think that would be useful in the beginning.

Done and committed to the branch

Issue #17, opened 2020-02-29
use workload identity


sirdarckcat:
you can have a GSA for a workload. might be useful for limiting access

sirdarckcat:
had to revert in https://github.com/google/kctf/pull/77

sirdarckcat:
Error was:
   https://github.com/google/kctf/runs/660654663?check_suite_focus=true#step:8:30

Also, this means that creating a cluster needs Owner permissions (Editor is not enough) since binding IAM permissions requires an additional permission (set IAM policy role, requires IAM Admin)

gree-gorey:
Is it an existing cluster?

I think that's because during the cluster creation one needs to use additional flag `--workload-pool` as I did in https://github.com/google/kctf/pull/75/commits/152e144d599d595e4ca0848abc47790eb649febe#diff-bb8a8efddd63e9c2d15b46c10737fbfaR29

I will try to adjust that so it can add this pool to existing cluster.

gree-gorey:
Smth like
```
gcloud container clusters update ${CLUSTER_NAME} --workload-pool=${PROJECT}.svc.id.goog
```

should do. I just didn't account for existing clusters.

sirdarckcat:
@gree-gorey - I added you to the ACL of the project, could you make the change and then merge your change again plz? :)

sirdarckcat:
it's called `kctf-github-ci`

gree-gorey:
I think we can close this one? It is implemented for GCS; are there any other services in GCP that we connect to from k8s?

PR #75, opened 2020-05-08
switch from SA key to workload identity to connect to GCS fuse

gree-gorey:
Issue #17.

### DIFF
--- a/scripts/cluster/start.sh
+++ b/scripts/cluster/start.sh
@@ -59,10 +59,7 @@ fi
 
 KSA_NAME="gcsfuse-sa"
 
-if ! kubectl get serviceaccount/${KSA_NAME} --namespace kube-system; then
-  kubectl create serviceaccount --namespace kube-system ${KSA_NAME}
-fi
-
+kubectl create serviceaccount --namespace kube-system ${KSA_NAME} --dry-run=client -o yaml | kubectl apply -f -
 gcloud iam service-accounts add-iam-policy-binding --role roles/iam.workloadIdentityUser --member "serviceAccount:${PROJECT}.svc.id.goog[kube-system/${KSA_NAME}]" ${GSA_EMAIL}
 kubectl annotate serviceaccount --namespace kube-system ${KSA_NAME} iam.gke.io/gcp-service-account=${GSA_EMAIL} --overwrite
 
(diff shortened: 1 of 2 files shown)


scripts/cluster/start.sh
```diff
@@ -57,13 +57,15 @@ if [ -z "${GSA_EMAIL}" ]; then
   done
 fi
 
-if ! kubectl get secret/gcsfuse-secrets --namespace kube-system; then
-  KEY_PATH=$(mktemp -d)/key.json
-  gcloud iam service-accounts keys create "${KEY_PATH}" --iam-account "${GSA_EMAIL}"
-  kubectl create secret generic gcsfuse-secrets --from-file="${KEY_PATH}" --namespace kube-system
-  rm -rf $(dirname "${KEY_PATH}")
+KSA_NAME="${GSA_NAME}"
```
sroettger, line 60: Could the KSA_NAME be some fixed value like "gcsfuse_sa" so that we don't have to replace it in the yaml?

gree-gorey: Yes, that's certainly better. Fixed.

scripts/cluster/start.sh
```diff
@@ -57,13 +57,15 @@ if [ -z "${GSA_EMAIL}" ]; then
...
 
-if ! kubectl get secret/gcsfuse-secrets --namespace kube-system; then
-  KEY_PATH=$(mktemp -d)/key.json
-  gcloud iam service-accounts keys create "${KEY_PATH}" --iam-account "${GSA_EMAIL}"
-  kubectl create secret generic gcsfuse-secrets --from-file="${KEY_PATH}" --namespace kube-system
-  rm -rf $(dirname "${KEY_PATH}")
+KSA_NAME="gcsfuse-sa"
+
+if ! kubectl get serviceaccount/${KSA_NAME} --namespace kube-system; then
+  kubectl create serviceaccount --namespace kube-system ${KSA_NAME}
```
sirdarckcat, line 63: Is the if necessary? Can we use the apply format instead?

sirdarckcat: See the way the configmap is created (it pipes a create into an apply)

gree-gorey: Yes, you are right. Changed for apply.

sirdarckcat:
Looks good. Are we ok to merge it from your side?

gree-gorey:
> Looks good. Are we ok to merge it from your side?

Yes, we can merge.

Issue #74, opened 2020-05-07
GCS bucket ACLs for uniform access-control


gree-gorey:
`gsutil acl ch -u "${GSA_EMAIL}:O" "gs://${BUCKET_NAME}"` from `scripts/cluster/start.sh` is only working with Fine-grained access-control. If Uniform access-control enabled for bucket, this needs to be changed for:
```
gsutil iam ch serviceAccount:${GSA_EMAIL}:roles/storage.legacyBucketOwner gs://${BUCKET_NAME}/
```
Probably need to add check before choosing the command.

gree-gorey:
And also `legacyObjectOwner`:
```
gsutil iam ch serviceAccount:${GSA_EMAIL}:roles/storage.legacyObjectOwner gs://${BUCKET_NAME}/
```

sirdarckcat:
Oh, I think last time I only added Bucket owner, and didn't add Object owner.. hmm.. :-)

Nice catch!

gree-gorey:
Thanks :) Yes, the bucket owner has all the permissions except reading objects, I had to learn it myself the hard way :(

sirdarckcat:
Sweet!

PR #76, opened 2020-05-08
checking for uniform access-control enabled for bucket

gree-gorey:
Issue #74.

Unfortunately `gsutil uniformbucketlevelaccess get` doesn't return exit code 1 on False, hence grep.

### DIFF
--- a/scripts/cluster/start.sh
+++ b/scripts/cluster/start.sh
@@ -68,6 +68,13 @@ if ! gsutil du "gs://${BUCKET_NAME}/"; then
   gsutil mb -l eu "gs://${BUCKET_NAME}/"
 fi
 
+if gsutil uniformbucketlevelaccess get "gs://${BUCKET_NAME}" | grep -q "Enabled: True"; then
+  gsutil iam ch "serviceAccount:${GSA_EMAIL}:roles/storage.legacyBucketOwner" "gs://${BUCKET_NAME}"
+  gsutil iam ch "serviceAccount:${GSA_EMAIL}:roles/storage.legacyObjectOwner" "gs://${BUCKET_NAME}"
+else
+  gsutil acl ch -u "${GSA_EMAIL}:O" "gs://${BUCKET_NAME}"
+fi
+
 gsutil acl ch -u "${GSA_EMAIL}:O" "gs://${BUCKET_NAME}"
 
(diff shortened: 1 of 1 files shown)


scripts/cluster/start.sh
```diff
@@ -68,6 +68,13 @@ if ! gsutil du "gs://${BUCKET_NAME}/"; then
...
 fi
 
+if gsutil uniformbucketlevelaccess get "gs://${BUCKET_NAME}" | grep -q "Enabled: True"; then
+  gsutil iam ch "serviceAccount:${GSA_EMAIL}:roles/storage.legacyBucketOwner" "gs://${BUCKET_NAME}"
+  gsutil iam ch "serviceAccount:${GSA_EMAIL}:roles/storage.legacyObjectOwner" "gs://${BUCKET_NAME}"
+else
+  gsutil acl ch -u "${GSA_EMAIL}:O" "gs://${BUCKET_NAME}"
+fi
+
 gsutil acl ch -u "${GSA_EMAIL}:O" "gs://${BUCKET_NAME}"
```
sroettger, line 78: is this line a leftover? I.e. it's the same command as in the else clause?

gree-gorey: Right, removed it.

scripts/cluster/start.sh
```diff
@@ -68,7 +68,12 @@ if ! gsutil du "gs://${BUCKET_NAME}/"; then
   gsutil mb -l eu "gs://${BUCKET_NAME}/"
 fi
 
-gsutil acl ch -u "${GSA_EMAIL}:O" "gs://${BUCKET_NAME}"
+if gsutil uniformbucketlevelaccess get "gs://${BUCKET_NAME}" | grep -q "Enabled: True"; then
```
sirdarckcat, line 71: I had issues with this before, did you test in both cases this works? I couldn't get uniform bucket acls to work with gcsfuse, but maybe I did something wrong.

gree-gorey: I checked it with uniformbucketlevelaccess enabled and can confirm that it's working.

sirdarckcat:
Ok, cool! Is it ok to merge from your side?

gree-gorey:
> Ok, cool! Is it ok to merge from your side?

Yes, we can merge.

PR #79, opened 2020-05-11
switch from SA key to workload identity to connect to GCS fuse

gree-gorey:
Issue #17, now trying to merge with updated k8s cluster.


gree-gorey:
Updated cluster, now it has `workload-pool` configured.

PR #85, opened 2020-05-27
update walkthrough

sirdarckcat:
I went through the walkthrough and updated a few things that seemed not to work.

I also removed the mention of the network configuration from this, as it's quite complicated. I think this should be a different doc.

### DIFF
--- a/base/challenge-skeleton/config/advanced/network/network.yaml
+++ b/base/challenge-skeleton/config/advanced/network/network.yaml
@@ -15,11 +15,3 @@ apiVersion: "v1"
 kind: "Service"
 metadata:
   name: "chal"
-#spec:
-#  ports:
-#  - name: "netcat"
-#    protocol: "TCP"
-## This will be the public port
-#    port: 1
-## This is the port exposed by the pod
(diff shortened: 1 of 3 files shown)


sirdarckcat:
https://console.cloud.google.com/cloudshell/open?git_repo=https://github.com/google/kctf&tutorial=docs/google-cloud.md&shellonly=true&cloudshell_git_branch=sirdarckcat-patch-3

base/challenge-skeleton/config/advanced/network/network.yaml
```diff
@@ -15,11 +15,3 @@ apiVersion: "v1"
 kind: "Service"
 metadata:
   name: "chal"
-#spec:
```
sroettger, line 18: You don't think the example here is useful?

sirdarckcat: They are a bit confusing, since they are explaining how to do something that is not needed (map port 1 to 1337). I think we should have a README here instead (and in the other folder) explaining what docs to read and an example.

PR #96, opened 2020-05-28
fix pow.py healthcheck library

sirdarckcat:
this uses a different format now https://github.com/google/kctf/blob/master/base/nsjail-docker/files/proof_of_work/pow.py#L112


sroettger:
why didn't the checks catch this?

sirdarckcat:
> why didn't the checks catch this?

https://github.com/google/kctf/issues/98

sirdarckcat:
https://github.com/google/kctf/issues/97

PR #101, opened 2020-06-09
Add HTTPS support

### LINKED ISSUE #78: TLS support for web tasks
Currently web tasks are http, but some tasks will require https. Setting this up atm is kindof complicated.

We need to somehow pass down to the task its own hostname, so it can request a certificate at runtime, or we need to do this at config time so we can request a certificate through k8s.

sirdarckcat:
Closes #78

This automatically provisions a certificate and builds a HTTPS load balancer.

Unfortunately this means we end up with 2 load balancers for the same task (2x the cost).


sirdarckcat:
live at https://apache-others.kctf-testing.kctf.dev/nodejs/foo

sroettger:
what happens if expose=false, https=true?

sirdarckcat:
can I haz approve? :P

sirdarckcat:
gree-gorey would you mind TAL? :) stephen is probably busy with kids :P

gree-gorey:
IIUC, the problem with two LB is that Managed certs in GKE are only available for Ingress. But we can't change all public challenges to use Ingress instead of just TCP LB, because some of them are not using http protocol -- is that correct?

So, maybe in that case we would better specify challenge type in config, say smth like `PROTOCOL=http|tcp`, and for http we would create *only* Ingress, for others we would create *only* network LB as before. (Btw, I should adjust my PR with DNS so that it annotate Ingresses as well because right now it only annotates LB).)

What about challenges that would want to use TLS with TCP? They can't use Ingress. Or we don't expect such things?

samples/kctf-conf/base/Makefile
```diff
@@ -19,7 +19,7 @@
 # See the License for the specific language governing permissions and
```
gree-gorey, line 19: This is the same exact file as `base/Makefile`, right? Why it's not a symlink like in `kctf/base/challenge-skeleton/Makefile`? Should I also duplicate changes in two Makefiles in my PR?

base/Makefile
```diff
@@ -149,7 +155,40 @@ clean:
...
+    serviceName: "chal"
+    servicePort: 1
+endef
+
+export CERTIFICATE_TEMPLATE
+export INGRESS_TEMPLATE
+
+.https: .cluster-config
+ifeq ($(HTTPS), true)
+	(echo "$${CERTIFICATE_TEMPLATE}"; echo '    -' "$(CHALLENGE_NAME).$(DOMAIN_NAME)") | kubectl apply -f -
```
gree-gorey, line 185: This command would fail if DOMAIN_NAME is not set. Should we check and inform user to save her/him troubleshooting time? (I need to do the same with DNS btw :)

samples/kctf-conf/base/Makefile
sirdarckcat: oh, yea.. /samples/ is the equivalent of what an "upgrade" in infrastructure would be. this means, every CL needs to update /samples/, but we always forget :(. we can't make it a symlink because in "real life" users wouldn't do this.

the way I usually do the "upgrade" is that I delete the `kctf-conf/base` directory and then run setup-chal-dir, which recreates it from "head".

sirdarckcat:
> IIUC, the problem with two LB is that Managed certs in GKE are only available for Ingress. But we can't change all public challenges to use Ingress instead of just TCP LB, because some of them are not using http protocol -- is that correct?

Right, Ingress is only HTTPS.

> So, maybe in that case we would better specify challenge type in config, say smth like `PROTOCOL=http|tcp`, and for http we would create _only_ Ingress, for others we would create _only_ network LB as before. (Btw, I should adjust my PR with DNS so that it annotate Ingresses as well because right now it only annotates LB).)

Makes sense. FWIW, Ingress has to point to a Service, so I'm currently pointing to the LoadBalancer Service, if we don't create the LB, then we will do a NodePort instead. Alternatively we could annotate it as internal https://cloud.google.com/kubernetes-engine/docs/how-to/internal-load-balancing

> What about challenges that would want to use TLS with TCP? They can't use Ingress. Or we don't expect such things?

I think TLS+TCP would be the "responsibility" of the challenge author to setup. I think cases when this might be necessary (SMTP+STARTTLS or just encrypting traffic), it would be easier for the task author to provide the certificate or CA of the key being used, since the players would need to do the cert checks manually anyway.

base/Makefile
sirdarckcat: good point, I'll update this CL to be based on yours once you submit yours

sirdarckcat:
If you have a CTF tasks with 2 components (eg, 2 ports - say port 80/443 and port 22), then you can't easily make it happen with DNS. Maybe instead of letting the user select HTTPS or TCP, we can leave this as `HTTPS=true/false` and that just creates *another* DNS name (say, `${challenge-name}-web.${domain-name}` and ${challenge-name}-tcp.${domain-name}`). WDYT?

gree-gorey:
> If you have a CTF tasks with 2 components (eg, 2 ports - say port 80/443 and port 22), then you can't easily make it happen with DNS. Maybe instead of letting the user select HTTPS or TCP, we can leave this as `HTTPS=true/false` and that just creates _another_ DNS name (say, `${challenge-name}-web.${domain-name}` and `${challenge-name}-tcp.${domain-name}`). WDYT?

I see. Yes, with current setup I think that would be the easiest way.

But ultimately I think would be better to assign each port a type of protocol it serves. Now it would be probably difficult to do with `.conf` files but when we migrate to CRD we could do smth like:

```yaml
...
  spec:
    ...
    ports:
    - port: 22
      protocol: "TCP"
    - port: 443
      protocol: "HTTP"
      https: true
```

Then we would create Ingress for http ports, and tcp LB for tcp ports.

sirdarckcat:
@googlebot I fixed it.

sirdarckcat:
@gree-gorey makes sense, let's do that on the CRD version! =)

I'm currently trying to get a demo running for this version of the code (got some IAM issues we need to resolve for me to test it see cl/318991759).

sirdarckcat:
one issue with my naming proposal, is that we will end up with:

`web-asm-web.domainname.com`

I'm not super thrilled by that (double `-web`), maybe it should be `-ssl` instead or `-https`.

gree-gorey:
> one issue with my naming proposal, is that we will end up with:
> 
> `web-asm-web.domainname.com`
> 
> I'm not super thrilled by that (double `-web`), maybe it should be `-ssl` instead or `-https`.

Right, `-http` sounds better.

gree-gorey:
`samples/bash/config/chal.conf` still has `PROTOCOL=tcp`

samples/kctf-conf/base/Makefile
```diff
@@ -149,14 +150,54 @@ clean:
...
+ifeq ($(HTTPS), true)
+ifneq ($(DNS), true)
+	$(error 'HTTPS=true but DNS=false. Need DNS to get certificate.')
+endif
+ifeq ($(DOMAIN_NAME),)
+        $(error 'HTTPS=true but DOMAIN_NAME is not set.')
+endif
+	(echo "$${CERTIFICATE_TEMPLATE}"; echo '    -' "$(CHALLENGE_NAME).$(DOMAIN_NAME)") | kubectl apply -f -
+	echo "$${INGRESS_TEMPLATE}" | kubectl apply -f -
+	kubectl annotate "ingress/https" "external-dns.alpha.kubernetes.io/hostname=${CHALLENGE_NAME}-http.${DOMAIN_NAME}" --overwrite
```
gree-gorey, line 188: btw, you don't need to annotate Ingresses: DNS entry is updated for them automatically using host form spec

sirdarckcat: hmm, but where does it get the host name from?

gree-gorey: From `spec.rules.host` within Ingress. If it's unspecified like now in Makefile it means that it matches any `*` hostname.

For dns-controller to be able to see domain we need to do:

```yaml
spec:
  rules:
  - host: ${CHALLENGE_NAME}-http.${DOMAIN_NAME}"
    http:
      paths:
      - backend:
          serviceName: "chal"
          servicePort: 1
```

sirdarckcat: hmm, does this mean we maybe would only need a single Ingress, and the route stuff based on the hostname to different ports?

say if your challenge has 3 containers, or 3 ports, we could expose 8080 on `${CHALLENGE_NAME}-8080.${DOMAIN_NAME}` and 8081 on `${CHALLENGE_NAME}-8080.${DOMAIN_NAME}` and so forth. Always using port 80/443, but mapping to different internal services.

This would probably save money, because public IPs and LoadBalancers are expensive, so having a single one per challenge is better (it would actually be better if we could have a single TCP load balancer for the whole CTF and just different port numbers, and then a single HTTPS load balancer and different hostnames).

sirdarckcat: hmm seems that this isn't possible to do cross-namespace ingress https://github.com/kubernetes/kubernetes/issues/17088

I guess we can still do this for multiple HTTP ports on a single task, though.

gree-gorey: Yes, for using it for multiple ports is a good idea.

sirdarckcat:
finally https://apache-others-http.kctf-testing.kctf.dev/nodejs/foo

samples/kctf-conf/base/Makefile
sirdarckcat: ok, since the current `chal.conf` model (with booleans) doesn't know of the multiple ports, I think we should move this idea to CRD. I'll file a bug for this.

sirdarckcat: https://github.com/google/kctf/issues/112

PR #103, opened 2020-06-20
add external DNS controller to automatically manage DNS records

gree-gorey:
[external-dns](https://github.com/kubernetes-sigs/external-dns) controller is setup for the cluster in the start.sh when `--domain-name` is specified, e.g.:
```console
$ kctf-config-create --chal-dir ~/kctf-demo --project  test-project --domain-name dns-test.kctf --start-cluster
```
It will watch for the annotations in Service of type LB, or for Ingresses to create DNS entries. E.g. if you create this LB:
```yaml
apiVersion: v1
kind: Service
metadata:
  name: nginx
  annotations:
    external-dns.alpha.kubernetes.io/hostname: my-chal.dns-test.kctf.
spec:
  type: LoadBalancer
# ...
```
then it will create DNS entry `my-chal.dns-test.kctf.`. Test with:
```console
$ dig +short @ns-cloud-d1.googledomains.com. my-chal.dns-test.kctf.
<LB public IP>
```

One thing for discussion: there is a parameter:
```bash
--policy=upsert-only # would prevent ExternalDNS from deleting any records, omit to enable full synchronization
```
which I disabled to allow full synchronization. We can choose upsert-only if needed.


config/external-dns.yaml
```diff
@@ -0,0 +1,74 @@
...
+      app: external-dns
+  template:
+    metadata:
+      labels:
+        app: external-dns
+    spec:
+      serviceAccountName: external-dns-sa
+      containers:
+      - name: external-dns
+        image: registry.opensource.zalan.do/teapot/external-dns:latest
```
sirdarckcat, line 61: is this opensource?

gree-gorey: Yes, the code is here: https://github.com/kubernetes-sigs/external-dns.
As it turned out, they have also pushed images to gcr.io, so I changed the image manifest to pull from GCR (but in the documentation they are referring to Zalando registry which may be just outdated documentation).

gree-gorey:
> nice! :)
> 
> is it possible to automatically add the annotation with the project name?

We probably should do it somewhere in our scripts, probably in Makefile when we are doing `kubectl apply -k config/advanced/network`, just add `kubectl annotate svc chal external-dns.alpha.kubernetes.io/hostname=${CHALLENGE_NAME}.${DOMAIN_NAME}`?
Or, somewhere in `config/advanced/network`.

sirdarckcat:
yea, lets do it on the Makefile - we already do it for HTTPS so maybe we can do it for DNS too. Maybe it can be a setting in `chal.conf`? (`DNS=true` by default?)

gree-gorey:
Added automatic annotaion.

scripts/cluster/start.sh
```diff
@@ -82,3 +82,36 @@ kubectl apply -f "${DIR}/config/daemon.yaml"
 kubectl apply -f "${DIR}/config/network-policy.yaml"
 kubectl apply -f "${DIR}/config/allow-dns.yaml"
 kubectl patch ServiceAccount default --patch "automountServiceAccountToken: false"
+
+# Cloud DNS
```
sirdarckcat, line 86: could you delete https://github.com/google/kctf/blob/master/scripts/setup/dns.sh

scripts/cluster/start.sh
```diff
@@ -82,3 +82,36 @@ kubectl apply -f "${DIR}/config/daemon.yaml"
...
+
+if [ ! -z "${DOMAIN_NAME}" ]
+then
+  DNS_ZONE=$(gcloud dns managed-zones list --filter "dns_name:${DOMAIN_NAME}" --format 'get(name)')
+  if [ -z "${DNS_ZONE}" ]; then
+    gcloud dns managed-zones create "${CLUSTER_NAME}-dns-zone" --description "DNS Zone for ${DOMAIN_NAME}" --dns-name="${DOMAIN_NAME}."
+  fi
+
+  GSA_NAME="kctf-cloud-dns"
```
sirdarckcat, line 95: could you change the name of the env var? maybe DNS_GSA_NAME? (and same for GCS_GSA_NAME)? so it's clear they are gonna be different

scripts/cluster/start.sh
```diff
@@ -82,3 +82,36 @@ kubectl apply -f "${DIR}/config/daemon.yaml"
...
+  GSA_EMAIL=$(gcloud iam service-accounts list --filter "email=${GSA_NAME}@${PROJECT}.iam.gserviceaccount.com" --format 'get(email)' || true)
```
sirdarckcat, line 96: same as before, DNS_GSA_EMAIL

scripts/cluster/start.sh
```diff
@@ -82,3 +82,36 @@ kubectl apply -f "${DIR}/config/daemon.yaml"
...
+  if [ -z "${GSA_EMAIL}" ]; then
+    gcloud iam service-accounts create "${GSA_NAME}" --description "kCTF Cloud DNS service account ${CLUSTER_NAME} ${ZONE}" --display-name "kCTF Cloud DNS ${CLUSTER_NAME} ${ZONE}"
+    GSA_EMAIL=$(gcloud iam service-accounts list --filter "email=${GSA_NAME}@${PROJECT}.iam.gserviceaccount.com" --format 'get(email)')
+    while [ -z "${GSA_EMAIL}" ]; do
+      sleep 1
+      GSA_EMAIL=$(gcloud iam service-accounts list --filter "email=${GSA_NAME}@${PROJECT}.iam.gserviceaccount.com" --format 'get(email)')
+    done
+  fi
+
+  KSA_NAME="external-dns-sa"
```
sirdarckcat, line 107: and here, DNS_KSA_NAME =)

config/external-dns.yaml
```diff
@@ -0,0 +1,74 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRole
+metadata:
+  name: external-dns-viewer
+rules:
+- apiGroups: [""]
+  resources: ["services","endpoints","pods"]
```
sirdarckcat, line 20: why isn't this `services,endpoints,podes,nodes`?

gree-gorey: That's a good question :) I copied it from their documentation, but I believe they just had historical reasons for that.

PR #105, opened 2020-06-23
Update local-testing.md

goedi02:
Updates based on Alicia's feedback.

### DIFF
--- a/docs/local-testing.md
+++ b/docs/local-testing.md
@@ -47,10 +47,10 @@ This will take a bit longer the first time it is run, as it has to build a chroo
 cd $CHALDIR/test-1
 make test-docker
 sudo apt-get install -y netcat
-nc 127.0.0.1 external_port
+nc 127.0.0.1 [external_port]
 ```
 
-When `make test-docker` runs, you will see `0.0.0.0:external_port->1337/tcp` in the terminal.
+When `make test-docker` runs, you will see `0.0.0.0:[external_port]->1337/tcp` in the terminal.
 
 If all went well, you should have a shell inside an nsjail bash (use Ctrl+C to exit):
 ```
--- a/docs/local-testing.md
+++ b/docs/local-testing.md
@@ -105,7 +105,7 @@ Use the following command to connect to the challenge:
 ```
 kubectl port-forward --namespace=[name_of_the_challenge] deployment/chal :1337 &
 ```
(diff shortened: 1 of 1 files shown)


docs/local-testing.md
```diff
@@ -86,10 +101,14 @@ This deploys the challenge and healthcheck to the local Kubernetes cluster.
 
 ### Connect to the challenge
 
+Use the following command to connect to the challenge:
 ```
-kubectl port-forward deployment/test-1 :1337 &
+kubectl port-forward --namespace=[name_of_the_challenge] deployment/chal :1337 &
 ```
 
+Example: If you ran the challenge sample in the `samples/apache-php` folder, specify `apache-php` as the namespace. After forwarding completes, you can access `127.0.0.1:external_port` from your browser and should see a page similar to this one:
```
aliciafmachado, line 109: I think this part would be better like this:
```suggestion
Example: If you ran the challenge sample in the `samples/apache-php` folder, specify `apache-php` as the namespace. After you run the port-forward command, it will be shown in the command line from which external port the challenge is being forwarded. With that information, you can access `127.0.0.1:external_port` from your browser and should see a page similar to this one: 
```

docs/local-testing.md
```diff
@@ -47,10 +47,19 @@ This will take a bit longer the first time it is run, as it has to build a chroo
 cd $CHALDIR/test-1
 make test-docker
 sudo apt-get install -y netcat
-nc 127.0.0.1 1337
+nc 127.0.0.1 external_port
```
aliciafmachado, line 50: What do you think about changing external_port to [external_port], the same way as [name_of_the_challenge]?

docs/local-testing.md
```diff
@@ -47,10 +47,19 @@ This will take a bit longer the first time it is run, as it has to build a chroo
...
 ```
 
-If all went well, you should have a shell inside an nsjail bash. If you don't, there might be some issues with your system (support for nsjail, Docker etc.). See the [debugging instructions](#errors-with-docker) further down on this page.
+When `make test-docker` runs, you will see `0.0.0.0:external_port->1337/tcp` in the terminal.
```
aliciafmachado, line 53: And if we change to [external_port], should we also change here?

sirdarckcat:
thanks!

PR #115, opened 2020-07-01
Delete extra files from apache-others

sirdarckcat:
In the migration to the new chal directory layout we added a bunch of files to the wrong directory.


sirdarckcat:
stephen is still half asleep I think :P

PR #117, opened 2020-07-17
Allow users to push images themselves

sroettger:
In some cases, it's not possible/desirable to create a docker image on the local machine in which case we can't push it from the main Makefile.
One example is if we want to create a windows docker image, we need to build and push it from a windows machine.

With this case, if the challenge/Makefile returns the full path to the image, i.e. eu.gcr.io/foo/bar:123456 instead of just the hash, the Makefile will assume it's already pushed.

### DIFF
--- a/base/Makefile
+++ b/base/Makefile
@@ -27,18 +27,20 @@ SHELL := bash
 
 PUSH_TARGET="REMOTE"
 
-REGISTRY:=CONFIGMISSING
-PROJECT:=CONFIGMISSING
-CLUSTER_NAME:=CONFIGMISSING
-ZONE:=CONFIGMISSING
+export REGISTRY:=CONFIGMISSING
+export PROJECT:=CONFIGMISSING
+export CLUSTER_NAME:=CONFIGMISSING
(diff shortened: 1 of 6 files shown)


sroettger:
There's also a drive by commit to remove default modes, but let me know if I should create a separate pull request for that.

base/Makefile
```diff
@@ -27,18 +27,20 @@ SHELL := bash
 
 PUSH_TARGET="REMOTE"
 
-REGISTRY:=CONFIGMISSING
-PROJECT:=CONFIGMISSING
-CLUSTER_NAME:=CONFIGMISSING
-ZONE:=CONFIGMISSING
+export REGISTRY:=CONFIGMISSING
```
sirdarckcat, line 30: what does this do?

sirdarckcat (approved):

base/Makefile
sroettger: it makes it available in the submake. That's how I pass to the challenge Makefile to which registry to push to

sirdarckcat: cool

PR #118, opened 2020-07-23
Beta dev

aliciafmachado:
Hi! I made the PR.

So, again, just to remind, the files that I edited are:

1. kctf-operator/pkg/apis/kctf/v1alpha1/challenge_types.go

This one is responsible for generating the CRDs.

2. The whole folder: kctf-operator/pkg/controller/challenge/.

There we have what the operator does every time a CR is applied - the reconcile function. So, the main file is challenge_controller.go, which contains the reconcile function itself. It basically stayed the same way as the boilerplate code, but I added the calls of these functions, which are contained in the others files: SetDefaultValues, deploymentForChallenge, serviceForChallenge and CheckConfigurations.

The other files were created so that the code stays more organised. The challenge_default has functions that return default values that can't be set by kubebuilder. The challenge_deployment contains the deployment. The challenge_service contains the service. Finally, the challenge_check is always called by the reconcile function (which is different from the deployment/service creation, which aren't called every time), which checks the configurations and make changes if necessary.

There are still some things to be done, and, thus, I left some TODOs in these places, so you can know if I didn't put something important or I simply didn't do it yet.


kctf-operator/pkg/apis/kctf/v1alpha1/challenge_types.go
```diff
@@ -0,0 +1,130 @@
...
+
+	// +kubebuilder:default:=true
+	Public bool `json:"public,omitempty"`
+
+	// +kubebuilder:default:=false
+	Dns bool `json:"dns,omitempty"`
+
+	// By default, one port is set with default values
+	// +kubebuilder:validation:Optional
+	Ports []corev1.ServicePort `json:"ports,omitempty"`
```
gree-gorey: Have you tried set port protocol to "HTTP"?
This is our "invention" and this string actually not supported by native k8s service: https://github.com/kubernetes/api/blob/master/core/v1/types.go#L4152

But I don't see that it's enforced anywhere so probably that should be fine. Just make sure to test that it works

kctf-operator/pkg/controller/challenge/challenge_check.go
```diff
@@ -0,0 +1,84 @@
...
+func CheckAutoscaling(challenge *kctfv1alpha1.Challenge, deployment *appsv1.Deployment) bool {
+	// Flag to say if there was any change done
+	change := false
+
+	if challenge.Spec.Deployed == true {
+
+		minRep := challenge.Spec.Autoscaling.MinReplicas
+		maxRep := challenge.Spec.Autoscaling.MaxReplicas
+
+		if *deployment.Spec.Replicas < minRep {
```
gree-gorey: We don't have to adjust Deployment replicas by Operator.
What we need is to create a HorizontalPodAutoscaler object based on these values, see example here: https://github.com/google/kctf/blob/alpha/samples/apache-php/config/advanced/deployment/autoscaling.yaml

kctf-operator/pkg/controller/challenge/challenge_check.go
```diff
@@ -0,0 +1,84 @@
...
+
+import (
+	kctfv1alpha1 "github.com/google/kctf/pkg/apis/kctf/v1alpha1"
+	appsv1 "k8s.io/api/apps/v1"
+)
+
+func CheckDeployed(challenge *kctfv1alpha1.Challenge, deployment *appsv1.Deployment) bool {
+	// First, check if the challenge is ready and, if not, set replicas to 0
+	if challenge.Spec.Deployed == false && *deployment.Spec.Replicas != 0 {
+		var numReplicas int32 = 0
```
gree-gorey: Also we should ensure here that HorizontalPodAutoscaler is disabled.
Because if it's enabled then it will increase replicas back.
You don't have to do it now, but make sure you add TODO for it

kctf-operator/build/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
```
sroettger, line 1: in other places, we use ubuntu:20.04 as the base image. Would that work here as a drop in replacement?
Don't worry if it doesn't. It would just reduce the amount of data that people have to download during setup.
@sirdarckcat I think there are a few other places where we use non-ubuntu base images. Do you think it makes sense to try to change those?

kctf-operator/pkg/controller/challenge/challenge_check.go
sroettger: I think it's a bit unexpected that the Check functions modify the arguments. Maybe we could move this code to the reconcile loop instead?

kctf-operator/pkg/controller/challenge/challenge_controller.go
```diff
@@ -0,0 +1,183 @@
...
+			return reconcile.Result{}, err
+		}
+
+		// Service deleted successfully - return and requeue
+		return reconcile.Result{}, err
+	}
+
+	// Ensure that the configurations in the CR are followed - Checks done everytime the CR is updated
+	// change says if something in the configurations was different from what was found in the deploymeny
+	change := CheckConfigurations(challenge, found)
```
sroettger: I don't fully understand what the Check is doing. At first I thought it would just check if all values in the config are valid but then I don't understand why it's happening after deployments etc were created.

kctf-operator/pkg/controller/challenge/challenge_default.go
```diff
@@ -0,0 +1,96 @@
...
+		Items: []corev1.PersistentVolumeClaim{
+			corev1.PersistentVolumeClaim{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: chal.Name,
+				},
+				Spec: corev1.PersistentVolumeClaimSpec{
+					Resources: corev1.ResourceRequirements{
+						Requests: corev1.ResourceList{
+							// Set 1Gi for the user
+							"storage": *resource.NewQuantity(1*1024*1024*1024*1024, resource.BinarySI),
```
sroettger: Isn't this 1Ti instead of 1Gi as the comment says?
Maybe you could use ParseQuantity here instead and we wouldn't need the comment anymore

kctf-operator/pkg/apis/kctf/v1alpha1/challenge_types.go
gree-gorey: I tested it and setting `protocol: "HTTP"` does work; but it doesn't work if you not specifying `port:` value. And `port` doesn't make sense with HTTP because it will be 80/443 anyway.
So maybe we better introduce our own object for that instead of using corev1.ServicePort?
It would be the same as corev1.ServicePort but it would have `port` as optional and allow `"HTTP"` value in protocol. Because using the standard object might confuse users (they can be surprised by usage of "HTTP" inside corev1.ServicePort)

kctf-operator/pkg/controller/challenge/challenge_default.go
aliciafmachado: Yes, it's true, I put it wrong. I will have a look at ParseQuantity and fix this.

kctf-operator/pkg/controller/challenge/challenge_controller.go
aliciafmachado: When I took the example of an operator, what we had in the challenge-controller was the deployment and, after that, a code that checked if the quantity of replicas was right. What I have done was to put the deployment and this code that checks if the configurations are well set in other files. Also, I created the service and created a function that set defaults values in variables that can't be set to default in CRD.

The deployment is created only once and the service as well if the variable public doesn't change. The only call that it's in all reconcile function calls is the CheckConfigurations, which changes the deployment/service if the author changes something in the CR. The other function that set default values is also called in all reconcile calls, because, when the author changes the CR, we have to pass them again.

kctf-operator/build/Dockerfile
gree-gorey: Just a comment: users don't have to download this. This image will be built and tagged by us and put in some public registry (gcr.io?) and then specified in operator deployment.

So only kctf-cluster admins will set that image to deploy Operator into the cluster. Then this image will be pulled by kctf-cluster nodes to run operator inside cluster

kctf-operator/pkg/controller/challenge/challenge_check.go
aliciafmachado: Yes, we could move to it, I just created other file because maybe the code would be too big in challenge-controller and, then, it would be harder for someone that wants to make changes in this code or simply understand it.

Maybe we could also change the world "check", so it's clearer to who reads the code.

kctf-operator/build/Dockerfile
aliciafmachado: I will see this and test if it works.

sirdarckcat: It still maybe helps for development to have a single base image. Is there a reason to use redhat here?

kctf-operator/pkg/controller/challenge/challenge_check.go
sirdarckcat: changing the name sgtm :)

kctf-operator/pkg/controller/challenge/challenge_controller.go
sirdarckcat: @sroettger would changing from check to a better name that describes what it does fix this concern?

kctf-operator/pkg/apis/kctf/v1alpha1/challenge_types.go
sirdarckcat: +1

kctf-operator/build/Dockerfile
sirdarckcat: oh, but maybe this is from the SDK, so maybe we can leave this for now?

kctf-operator/pkg/apis/kctf/v1alpha1/challenge_types.go
aliciafmachado: I didn't know that it's not supported by k8s. How do we pass currently a port protocol "HTTP" then?

kctf-operator/deploy/crds/kctf.dev_v1alpha1_challenge_cr_example_1.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: kctf.dev/v1alpha1
+kind: Challenge
+metadata:
+  name: mychal
```
sirdarckcat, line 4: is the namespace defined here?

kctf-operator/pkg/controller/challenge/challenge_check.go
aliciafmachado: So we just have to store the HorizontalPodAutoscaler and do nothing? Or should we pass to the deployment?

kctf-operator/deploy/operator.yaml
```diff
@@ -0,0 +1,33 @@
...
+  template:
+    metadata:
+      labels:
+        name: kctf-operator
+    spec:
+      serviceAccountName: kctf-operator
+      containers:
+        - name: kctf-operator
+          # Replace this with the built image name
+          image: REPLACE_IMAGE
```
sirdarckcat, line 20: who has to replace it?

kctf-operator/pkg/controller/challenge/challenge_check.go
gree-gorey: Yeah, I think the problem here is with naming.
So basically what happens here is these functions check the pushed config ("desired state") and actual objects ("actual state") and make sure that actual state matches the desired state.

maybe smth like `ensureDeployed`? (or other better name)

btw, does the function have to be exported? They are all in the same package, right?

kctf-operator/pkg/apis/kctf/v1alpha1/challenge_types.go
gree-gorey: We don't :) It's a change that comes with CRD migration.

What we have to do is:
* if the port is of type TCP, we create a LoadBalancer (just pass the values to the Service)
* if the port is of type HTTP we create an Ingress

kctf-operator/pkg/controller/challenge/challenge_controller.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+
+	// Watch for changes to primary resource Challenge
+	err = c.Watch(&source.Kind{Type: &kctfv1alpha1.Challenge{}}, &handler.EnqueueRequestForObject{})
+
+	if err != nil {
+		return err
+	}
+
+	// TODO(user): Modify this to be the types you create that are owned by the primary resource
```
sirdarckcat: do we need to do anything here?

kctf-operator/pkg/controller/challenge/challenge_check.go
gree-gorey: No, we don't have to change anything in the Deployment. Just need to create HPA object.

sirdarckcat (commented):
Thanks! Maybe for future code reviews we can do them in 3 rounds? Rather than all at once

kctf-operator/deploy/crds/kctf.dev_v1alpha1_challenge_cr_example_1.yaml
aliciafmachado: So, actually I was checking about the namespaces today. If we declare here the namespace, it will be the one used for the deployment and etc. Also, the operator can watch all namespaces. But I don't think it's ideal to declare the namespace there, right? And, just to confirm, the operator should use as default the name of the challenge as namespace, right?

The only thing that may be a problem now is that the namespace that we put there must already exist (at least with this code). I think it's the operator's duty to create a namespace for each challenge, right? I will do that next, but still need to figure out how to create one using Go.

kctf-operator/pkg/controller/challenge/challenge_controller.go
aliciafmachado: This TODO came automatically with the file, but until now, I didn't need to change anything there, so it might not be needed to modify it at all.

kctf-operator/deploy/operator.yaml
aliciafmachado: Actually I didn't try to run inside the cluster yet because I was trying to code the most part and then test this way but I think this is used when you run it inside the cluster: https://v0-18-x.sdk.operatorframework.io/docs/golang/quickstart/#1-run-as-a-deployment-inside-the-cluster . The operator needs an image to run.

kctf-operator/build/Dockerfile
sroettger: > Just a comment: users don't have to download this. This image will be built and tagged by us and put in some public registry (gcr.io?) and then specified in operator deployment.

Ah, if users are not supposed to push it to their own registry then I think it doesn't matter :)

kctf-operator/pkg/controller/challenge/challenge_check.go
sroettger: +1, changing the name would resolve my confusion

kctf-operator/pkg/controller/challenge/challenge_controller.go
sroettger: as above, changing the name would resolve this

kctf-operator/pkg/controller/challenge/challenge_check.go
aliciafmachado: Ok, so I will replace all "check" for "ensure" if it's ok.

Yes, I left all these five files in the same package and, then, the functions from these files don't need to be exported since they are from the same package.

kctf-operator/pkg/apis/kctf/v1alpha1/challenge_types.go
aliciafmachado: Ok, that makes sense. I will create then a new type with only the fields that we discussed (port, targetPort, name and protocol) and then create an Ingress if the protocol is HTTP. Btw, where is the Ingress created so I know which fields need to be configured in the Ingress?

kctf-operator/pkg/controller/challenge/challenge_check.go
aliciafmachado: Actually what do you think about "verify"? It sounds better for me.

kctf-operator/pkg/apis/kctf/v1alpha1/challenge_types.go
gree-gorey: `INGRESS_TEMPLATE` in https://github.com/google/kctf/blob/alpha/base/Makefile#L164

kctf-operator/deploy/operator.yaml
gree-gorey: Yes, that's right. As mentioned somewhere above we would have to build & push the operator image to some public registry (e.g. gcr.io) and put here the value. Cluster admins don't have to change it.

But we should better do it using some sort of CI/CD, not manually

kctf-operator/pkg/controller/challenge/challenge_check.go
gree-gorey: verify has the same problem as check: it assumes that you only checking/verifying values using some sort of validation.

I think smth like apply/update/ensure are better verbs for these functions

kctf-operator/pkg/apis/kctf/v1alpha1/challenge_types.go
aliciafmachado: And where is it used? I was looking at the specs of Service but didn't find it easily.

kctf-operator/pkg/apis/kctf/v1alpha1/challenge_types.go
```diff
@@ -0,0 +1,146 @@
...
+
+	// Healthcheck checks if the challenge works
+	// If empty, healthcheck is not enabled by default
+	// +kubebuilder:validation:Optional
+	Healthcheck HealthcheckSpec `json:"healthcheck,omitempty"`
+
+	// Autoscaling features determine quantity of replicas and CPU utilization
+	// If empty, autoscaling is not enabled by default
+	// +kubebuilder:validation:Optional
+	HorizontalAutoscaling autoscalingv1.HorizontalPodAutoscaler `json:"horizontalAutoscaling,omitempty"`
```
gree-gorey: I would recommend rather use `HorizontalPodAutoscalerSpec` ([api ref](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#horizontalpodautoscalerspec-v1-autoscaling)) because users don't need to set other fields like metadata

But be aware that changing it from your own defined resource as before to standard HorizontalPodAutoscalerSpec you allow users to change `scaleTargetRef` which should point to the Deployment. So either Operator should set this field by default so that it points to our Deployment, or use our own defined `AutoscalingSpec` as before, you can decide which way is better

See how scaleTargetRef is set here:  https://github.com/google/kctf/blob/alpha/base/k8s/deployment/autoscaling.yaml#L19

kctf-operator/pkg/apis/kctf/v1alpha1/challenge_types.go
```diff
@@ -0,0 +1,146 @@
...
+
+	// +kubebuilder:default:=false
+	Enabled bool `json:"enabled,omitempty"`
+
+	// +kubebuilder:default:="healthcheck"
+	Container string `json:"container,omitempty"`
+}
+
+// Autoscaling Specifications
+type AutoscalingSpec struct {
```
gree-gorey: Leave only one of them: either AutoscalingSpec or HorizontalPodAutoscalerSpec (see my comment below for details).

kctf-operator/pkg/controller/challenge/challenge_update.go
```diff
@@ -0,0 +1,64 @@
...
+import (
+	kctfv1alpha1 "github.com/google/kctf/pkg/apis/kctf/v1alpha1"
+	appsv1 "k8s.io/api/apps/v1"
+)
+
+func UpdateDeployed(challenge *kctfv1alpha1.Challenge, deployment *appsv1.Deployment) bool {
+	// First, ensure if the challenge is ready and, if not, set replicas to 0
+	// TODO: check if horizontal autoscaling is enabled
+	if challenge.Spec.Deployed == false && *deployment.Spec.Replicas != 0 {
+		var numReplicas int32 = 0
```
gree-gorey, line 14: nit: I saw that others use helper functions that return a pointer, so that you can inline this. If you expect you would need a pointer to int32 often you might consider adding such a helper.

Example: https://github.com/elastic/cloud-on-k8s/blob/master/pkg/utils/pointer/numeric.go#L8

gree-gorey: (it's not blocking; you can add TODO if you decide to implement it)

kctf-operator/pkg/apis/kctf/v1alpha1/challenge_types.go
aliciafmachado: I thought I had erased it. I will fix that.

kctf-operator/pkg/apis/kctf/v1alpha1/challenge_types.go
```diff
@@ -0,0 +1,129 @@
...
+
+	// Healthcheck checks if the challenge works
+	// If empty, healthcheck is not enabled by default
+	// +kubebuilder:validation:Optional
+	Healthcheck HealthcheckSpec `json:"healthcheck,omitempty"`
+
+	// Autoscaling features determine quantity of replicas and CPU utilization
+	// If empty, autoscaling is not enabled by default
+	// +kubebuilder:validation:Optional
+	HorizontalAutoscaling autoscalingv1.HorizontalPodAutoscalerSpec `json:"autoscaling,omitempty"`
```
gree-gorey: nit: rename the key to `HorizontalPodAutoscalerSpec` to match the object inside it

kctf-operator/deploy/crds/kctf.dev_v1alpha1_challenge_cr_example_5.yaml
```diff
@@ -0,0 +1,14 @@
...
+metadata:
+  name: mychal5
+spec:
+  # Add fields here
+  imageTemplate: "memcached:1.4.36-alpine"
+  deployed: true
+  network:
+    public: false
+    dns: false
+  autoscaling:
```
gree-gorey: update the name of this field in the example

kctf-operator/pkg/controller/challenge/challenge_controller.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+
+	// Creates autoscaling object
+	// Checks if maxreplicas exists to see if it's enabled
+	// If enabled, it checks if it already exists
+	autoscalingFound := &autoscalingv1.HorizontalPodAutoscaler{}
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: challenge.Name,
+		Namespace: challenge.Namespace}, autoscalingFound)
+
+	if challenge.Spec.HorizontalAutoscaling.MaxReplicas != 0 && err != nil && errors.IsNotFound(err) {
```
gree-gorey: I was trying to understand what's wrong with these comparisons for a while now.

I think the solution for these would be to use pointers in challenge.Spec like this:

```
HorizontalPodAutoscalerSpec *autoscalingv1.HorizontalPodAutoscalerSpec
```

And then all you have to do is to check that it's not nil:
```
if challenge.Spec.HorizontalPodAutoscalerSpec != nil ...
```

The same goes for the comparisons inside SetDefaultValues():
```
if challenge.Spec.PodTemplate != nil {}
```

etc.

gree-gorey (approved):
LGTM

kctf-operator/deploy/operator.yaml
aliciafmachado: Yes, I will leave a TODO there so we can fix this in a next PR when I start to run inside the cluster

sroettger (approved):

sirdarckcat (approved):

PR #124, opened 2020-08-11
Operator initializer and finalizer for challenge

aliciafmachado:
Now the operator creates some initial features that are necessary for the kctf (apparmor and etc) and then it starts watching the challenges. 

I also added a finalizer that erases the namespace when the challenge is deleted. As we do that, I added a condition that only accepts namespaces equal the name and rejects a namespace equal default (should we add other namespaces that are forbidden?). 

It wasn't possible to add a initializer for the challenge to create the namespace because the namespace has to be created before the challenge is created.


aliciafmachado:
Wrong branch

PR #125, opened 2020-08-11
Operator initializer and challenge finalizer

aliciafmachado:
Now the operator creates some initial features that are necessary for the kctf (apparmor and etc) and then it starts watching the challenges.

I also added a finalizer that erases the namespace when the challenge is deleted. As we do that, I added a condition that only accepts namespaces equal the name and rejects a namespace equal default (should we add other namespaces that are forbidden?).

It wasn't possible to add a initializer for the challenge to create the namespace because the namespace has to be created before the challenge is created.


kctf-operator/pkg/resources/daemon-ctf.go
```diff
@@ -0,0 +1,91 @@
+package resources
+
+import (
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+func NewDaemonSetCtf() runtime.Object {
```
sroettger, line 10: Not that it's planned, but if we wanted to ship some binary files with the daemon set in the future, any idea how we could do this?

sroettger: Maybe we can provide these daemons as container images on gcr.io or so in the future

sroettger (approved):
Looks good to me.
What's the main motivation for moving the daemonsets etc from yaml files to code in the operator? I guess it will help with versioning and making sure we know the state of the cluster?

aliciafmachado:
I guess I didn't think much about that, but I thought it was a good idea since we would have all kubernetes configurations in the same place? But we could also leave the yaml files if you think it's better.

kctf-operator/pkg/resources/daemon-ctf.go
aliciafmachado: What do you mean by shipping some binary files? And how could we do it with yaml files?

sroettger: ah, I was thinking mainly that we might want to replace that shell command with a binary at some point. But yeah, for that we would have to build a docker image anyway.
On the command line, you can do something like `kubectl create configmap --from-file=` but a docker image would be better in that case anyway I think. So nevermind, doesn't sound like something we need to think about right now.

sroettger:
Yeah, I think it's nice to be able to install everything needed in the cluster with a single kubectl command and for versioning we'll only have to check the version of the operator.

sirdarckcat (approved):

PR #126, opened 2020-08-17
Creation of packages and full implementation of updates for services, autoscaling and deployment

aliciafmachado:
Most of the changes here were simply the creation of the packages and the implementation of updates (not necessarily creation or deletion, but if the author changes only a field and then the object needs to be updated).
Created too the structure for pow and the volumes but it's not implemented yet.
Deleted the finalizer: it will no longer be needed. We can clean the persistent volume with simply one configuration. It's not necessary to create a controller for this. 
Small problem: Realized that the problem of creating the CRD with the pod template still is a problem. But if we erase two problematic lines in the CRD, the problem is solved. The bad part is that we would always need to do this every time we want to change the structure of the CRD (which might not be such a problem but needs to be documented).
UPDATE: We might still need a finalizer or an additional operator since a standard operator for persistent volumes is not implemented for GCS.


kctf-operator/pkg/apis/kctf/v1alpha1/challenge_types.go
```diff
@@ -83,27 +83,27 @@ type ChallengeSpec struct {
...
+	// +optional
 	Healthcheck HealthcheckSpec `json:"healthcheck,omitempty"`
 
 	// Autoscaling features determine quantity of replicas and CPU utilization
 	// If empty, autoscaling is not enabled by default
-	// +kubebuilder:validation:Optional
+	// +optional
 	HorizontalPodAutoscalerSpec *HorizontalPodAutoscalerSpec `json:"horizontalPodAutoscalerSpec,omitempty"`
 
 	// PodTemplate is used to set the paths of sessions and uploads
```
gree-gorey: session and uploads are too specific for apache-php challenge. We should change the comment to smth more generic. Btw, we can use PodTemplate not only to customize volume paths but also for other fields.

kctf-operator/pkg/controller/challenge/autoscaling/functions.go
```diff
@@ -0,0 +1,47 @@
...
+
+	"github.com/go-logr/logr"
+	kctfv1alpha1 "github.com/google/kctf/pkg/apis/kctf/v1alpha1"
+	autoscalingv1 "k8s.io/api/autoscaling/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+func CreateAutoscaling(challenge *kctfv1alpha1.Challenge, client client.Client, scheme *runtime.Scheme,
```
gree-gorey: Usually it's better for readability to avoid double naming like `autoscaling.CreateAutoscaling()` and instead just use `autoscaling.Create()`, as we expect only one Create function inside the package autoscaling (the same applies for other packages).

kctf-operator/pkg/controller/challenge/autoscaling/horizontal-pod-autoscaler.go
```diff
@@ -0,0 +1,29 @@
+package autoscaling
+
+import (
+	kctfv1alpha1 "github.com/google/kctf/pkg/apis/kctf/v1alpha1"
+	autoscalingv1 "k8s.io/api/autoscaling/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func AutoscalingForChallenge(challenge *kctfv1alpha1.Challenge) *autoscalingv1.HorizontalPodAutoscaler {
```
gree-gorey: Also maybe rename it to smth like `generate()` for simplicity? Also, should it be imported/public?

kctf-operator/pkg/controller/challenge/challenge_controller.go
```diff
@@ -112,86 +110,33 @@ func (r *ReconcileChallenge) Reconcile(request reconcile.Request) (reconcile.Res
 
 	if IsNamespaceAcceptable(request.NamespacedName) {
```
gree-gorey: This is not a code from this PR, but still.
It's better for readability that you don't check the second time for `IsNamespaceAcceptable()`; instead return early from the first conditional like this:

```
if !IsNamespaceAcceptable(request.NamespacedName) {
  // ...
  return reconcile.Result{}, nil
}

// the rest of the code without second condition
return reconcile.Result{}, nil

```

kctf-operator/pkg/controller/challenge/deployment/deployment-with-healthcheck.go
```diff
@@ -0,0 +1,80 @@
...
+import (
+	kctfv1alpha1 "github.com/google/kctf/pkg/apis/kctf/v1alpha1"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	resource "k8s.io/apimachinery/pkg/api/resource"
+	intstr "k8s.io/apimachinery/pkg/util/intstr"
+)
+
+// Deployment with Healthcheck
+func deploymentWithHealthcheck(challenge *kctfv1alpha1.Challenge) *appsv1.Deployment {
```
gree-gorey: Again, just `withHealthcheck()` would be cleaner

kctf-operator/pkg/controller/challenge/update/update.go
```diff
@@ -0,0 +1,270 @@
...
+	return true
+}
+
+// Copy ports from one service to another
+func CopyPorts(found *corev1.Service, wanted *corev1.Service) {
+	found.Spec.Ports = []corev1.ServicePort{}
+	found.Spec.Ports = append(found.Spec.Ports, wanted.Spec.Ports...)
+}
+
+func UpdateNumReplicas(challenge *kctfv1alpha1.Challenge, currentReplicas *int32) bool {
```
gree-gorey: here and below: cleaner without Update word, just `update.NumReplicas()`

kctf-operator/pkg/utils/utils.go
```diff
@@ -1,12 +1,3 @@
 // Functions that are useful
 
 package utils
```
gree-gorey: Do we still need this file? Or do you want to keep it for future functions?

kctf-operator/pkg/controller/challenge/set/default.go
```diff
@@ -1,5 +1,5 @@
 // File that set values correctly and return default values that weren't specified
-package challenge
+package set
```
gree-gorey, line 2: same for functions in this package: without word Set is cleaner

kctf-operator/pkg/apis/kctf/v1alpha1/challenge_types.go
aliciafmachado: Yes, I'm aware of that, I will change the comment then!

kctf-operator/pkg/controller/challenge/autoscaling/horizontal-pod-autoscaler.go
aliciafmachado: This function should be public because of the way I'm checking if there was a change in the fields. I agree with the idea of changing the names!

kctf-operator/pkg/utils/utils.go
aliciafmachado: I created it to put functions that are common to different controllers (from the moment that we have multiple controllers). But currently we don't use it, so I can delete it.

kctf-operator/pkg/controller/challenge/update/update.go
aliciafmachado: The thing is that we only use this updateNumReplicas inside the package, so I think it might not be very meaningful to just put numReplicas or is it? (I will correct the initial of some functions that only have internal use)

kctf-operator/pkg/controller/challenge/set/default.go
aliciafmachado: Same thing as numReplicas, it makes sense for the last function which is the one used in other package but the three first ones are of internal usage. Do you think it's better to leave without the set (like setPersistentVolumeClaim)?

kctf-operator/pkg/controller/challenge/update/update.go
gree-gorey: yeah, I think you can leave it as it is

kctf-operator/pkg/controller/challenge/set/default.go
gree-gorey: same here, you can leave it

kctf-operator/pkg/controller/challenge/challenge_controller.go
```diff
@@ -108,37 +108,36 @@ func (r *ReconcileChallenge) Reconcile(request reconcile.Request) (reconcile.Res
 		r.client.Delete(ctx, challenge)
```
gree-gorey, line 108: I think you need `return reconcile.Result{}, nil` here for my suggested change to work? So that in case when we have invalid name we stop executing the fuction

aliciafmachado: That's right, forgot about that. Fixed!

gree-gorey (approved):
LGTM

PR #128, opened 2020-08-25
Adding volumes, configmaps, secrets and certificates and restructuring the code again

aliciafmachado:
Now, all the expected functionalities work: volumes, pow, deployment, service, ingress, certificate, secrets, etc. 
About the secrets, I decided to create them in the kube-system and then copy them to other namespaces.
I added as well two new fields in the CRD: domainName and replicas.
Also, made some changes in the structure of the code, by passing the update functions to each package, so we have always a set of functions for the packages, unless one/some of them is/are unnecessary: create, delete, generate, isEqual and Update.
But it's still necessary to: finish the implementation of the status and maybe add others status; change the scripts; and finally add documentation.


kctf-operator/pkg/apis/kctf/v1alpha1/challenge_types.go
```diff
@@ -102,27 +109,32 @@ type ChallengeSpec struct {
 	PodTemplate *corev1.PodTemplate `json:"podTemplate"`
 
 	// PersistentVolumeClaim are used to determine how much resources the author requires for its challenge
-	// Default value: size 1Gb
 	// +optional
-	PersistentVolumeClaims *corev1.PersistentVolumeClaimList `json:"persistentVolumeClaim,omitempty"`
+	Claims []string `json:"claims,omitempty"`
```
gree-gorey: I would prefer name "Volumes" here or "VolumeClaims", or even leave the old name. The reason is that authors probably don't know anything about k8s and the word "claims" just doesn't tell a person anything and can confuse people.

kctf-operator/pkg/apis/kctf/v1alpha1/challenge_types.go
```diff
@@ -102,27 +109,32 @@ type ChallengeSpec struct {
 	PodTemplate *corev1.PodTemplate `json:"podTemplate"`
 
 	// PersistentVolumeClaim are used to determine how much resources the author requires for its challenge
```
gree-gorey: Also we should change the comment because we don't specify the resources here; just provide the names for the volumes

gree-gorey (approved):
Overall LGTM, see my comments.

I think that using ConfigMap for powDifficulty is a bit of over-engineering but I understand that this is not smth you developed, you just migrated it from the Makefile. I think the problem is as simple as "propagating an integer from config into a container" and should be solved by adding the ENV variable into Pod. That would require though to change pow script to read the ENV instead of the file here: https://github.com/google/kctf/blob/alpha/base/nsjail-docker/files/proof_of_work/maybe_pow.sh#L21.

But that is probably better for separate Issue/PR. This one is working and that's okay 👍

sirdarckcat:
> Overall LGTM, see my comments.
> 
> I think that using ConfigMap for powDifficulty is a bit of over-engineering but I understand that this is not smth you developed, you just migrated it from the Makefile. I think the problem is as simple as "propagating an integer from config into a container" and should be solved by adding the ENV variable into Pod. That would require though to change pow script to read the ENV instead of the file here: https://github.com/google/kctf/blob/alpha/base/nsjail-docker/files/proof_of_work/maybe_pow.sh#L21.
> 
> But that is probably better for separate Issue/PR. This one is working and that's okay 👍

I think the only downside of that is that we need to restart the deployment for it to take effect?

kctf-operator/deploy/operator.yaml
```diff
@@ -17,7 +17,7 @@ spec:
       containers:
         - name: kctf-operator
           # TODO: Replace this with the built image name
-          image: quay.io/aliciafmachado/kctf-operator:v0.0.1
+          image: quay.io/aliciafmachado/kctf-operator:latest
```
sirdarckcat, line 20: we should eventually move this to one in Google Container Registry

sirdarckcat: not now, but just a reminder :)

kctf-operator/pkg/controller/challenge/autoscaling/functions.go
```diff
@@ -26,22 +35,63 @@ func Create(challenge *kctfv1alpha1.Challenge, client client.Client, scheme *run
...
-		log.Error(err, "Failed to delete Autoscaling")
+		log.Error(err, "Failed to delete Autoscaling", "Autoscaling name: ",
+			autoscalingFound.Name, " with namespace ", autoscalingFound.Namespace)
 		return false, err
 	}
 
 	return true, nil
 }
+
+func Update(challenge *kctfv1alpha1.Challenge, client client.Client, scheme *runtime.Scheme,
```
sirdarckcat, line 61: why are the other functions lowecase?

kctf-operator/pkg/controller/challenge/deployment/functions.go
```diff
@@ -57,9 +94,56 @@ func Create(challenge *kctfv1alpha1.Challenge, cl client.Client, scheme *runtime
...
-		return reconcile.Result{}, err
+		return false, err
 	}
 
 	// Deployment created successfully - return and requeue
-	return reconcile.Result{Requeue: true}, nil
+	return true, nil
+}
+
+func Update(challenge *kctfv1alpha1.Challenge, client client.Client, scheme *runtime.Scheme,
```
sirdarckcat, line 104: ditto

kctf-operator/pkg/controller/challenge/dns/functions.go
```diff
@@ -0,0 +1,111 @@
...
+	if err != nil {
+		log.Error(err, "Failed to delete Certificate", "Certificate name: ",
+			certificateFound.Name, " with namespace ", certificateFound.Namespace)
+		return false, err
+	}
+
+	return true, nil
+}
+
+func Update(challenge *kctfv1alpha1.Challenge, client client.Client, scheme *runtime.Scheme,
```
sirdarckcat, line 59: ++

kctf-operator/pkg/controller/challenge/dns/certificate.go
```diff
@@ -0,0 +1,23 @@
...
+
+func generate(challenge *kctfv1alpha1.Challenge) *netgkev1beta1.ManagedCertificate {
+	certificate := &netgkev1beta1.ManagedCertificate{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "kctf-certificate",
+			Namespace: challenge.Namespace,
+		},
+		Spec: netgkev1beta1.ManagedCertificateSpec{
+			Domains: []string{
+				challenge.Name + "." + challenge.Spec.Network.DomainName,
```
sirdarckcat, line 17: hmm is this configured per cluster? or per task.

if it's per task, then we probably don't need to concatenate.

sirdarckcat (approved):

kctf-operator/deploy/operator.yaml
aliciafmachado: actually, i wanted to ask: where to should we move it?

kctf-operator/pkg/controller/challenge/autoscaling/functions.go
aliciafmachado: In go, functions that are uppercase mean that they are exportable; so, the only function that I use outside the package is Update, which has uppercase for this reason. I only use the other ones inside the package, which is the reason for them being lowercase.

kctf-operator/pkg/controller/challenge/dns/certificate.go
aliciafmachado: It's per task (at least this was the way it was in the Makefile)

aliciafmachado: You mean concatenate the domains?

kctf-operator/deploy/operator.yaml
sirdarckcat: we need to create a project for it, I think we already have a kctf-docker one. we can talk about this in the chat.

kctf-operator/pkg/controller/challenge/autoscaling/functions.go
sirdarckcat: k thx

kctf-operator/pkg/controller/challenge/dns/certificate.go
sirdarckcat: we can chat about this later :)

PR #132, opened 2020-08-26
Always update chroot

sroettger:
Always update the chroot during nsjail build.

Once we have versioning, we will always use a particular version of the chroot.
On one hand this is undesirable since you will have an outdated user space (libc etc).
On the other hand, updating it might break long running challenges. For example a heap challenge being broken by new libc mitigations.

What do you think? I actually lean on not changing this atm :)

### DIFF
--- a/base/nsjail-docker/Dockerfile
+++ b/base/nsjail-docker/Dockerfile
@@ -24,6 +24,8 @@ RUN /usr/sbin/useradd --no-create-home -u 1000 user
 COPY --from=bin /usr/bin/nsjail /usr/bin/nsjail
 COPY --from=chroot /chroot /chroot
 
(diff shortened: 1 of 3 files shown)


sirdarckcat (approved):

sroettger:
@sirdarckcat did you read the comment? :)

sirdarckcat:
sorry! :)

yes, I think since alpha is deprecated, we can assume that everything runs on latest, so this shouldn't matter too much? so whatever helps in the short term for the CTFs that will be running with the alpha infra.

sroettger:
so I think it will be the better experience if the chroot doesn't change by default during the lifetime of a challenge.
And if users want to update it, they can do it themselves.

PR #134, opened 2020-09-02
Status and Integration with the rest of the code

aliciafmachado:
Changed the scripts/makefile/code so the operator is integrated with everything else. I also created two status: Health and Status. The first one says if the challenge is healthy and the second one says what's happening with the challenge's pods.


base/Makefile
```diff
@@ -322,55 +255,16 @@ ${CLUSTER_GEN}/remote-healthcheck-image: ${CLUSTER_GEN}/healthcheck-image-pushed
 	kubectl patch ServiceAccount default --patch "automountServiceAccountToken: false" >&2
 	mkdir -p ${CLUSTER_GEN} >&2
 
-.FORCE:
```
sroettger, line 325: hmm, I don't understand how this works if the .FORCE target is not there anymore. I.e. it's still used by the targets above?

kctf-operator/pkg/apis/kctf/v1alpha1/challenge_types.go
```diff
@@ -1,6 +1,6 @@
 // This file is responsible for generating CRD (Custom Resource Definition)
 // kubebuilder might be used to set: default values, optional fields and etc
-// +kubebuilder:validation:Required
+// +kubebuilder:validation:Optional
```
sroettger, line 3: does this make everything in this file optional?

base/Makefile
aliciafmachado: I forgot to delete it in the other targets, but, no, it's not used anymore. I will make a commit deleting it in the other targets.

kctf-operator/pkg/apis/kctf/v1alpha1/challenge_types.go
aliciafmachado: Yes, it does that, but you can specify some fields that are required. I left this way, because most of the fields are optional.

kctf-operator/pkg/controller/challenge/utils/utils.go
```diff
@@ -0,0 +1,42 @@
...
+
+	if err == nil {
+		domainName = configmap.Data["DOMAIN_NAME"]
+	}
+
+	return domainName
+}
+
+// Find index of the container with a specific name in a list of containers
+func Find_idx(name string, containers []corev1.Container) int {
```
sroettger: All the other functions are in camel case, should this be too?
Also maybe it could include the word "Container" to be more descriptive in what it does, e.g. IndexOfContainer()

sroettger (approved):
cool, looks good. Just a few comments.

kctf-operator/pkg/controller/challenge/utils/utils.go
aliciafmachado: Yes, sure, I will change to this name :)

base/Makefile
sroettger: what changed that it's not needed anymore?
For example in:
```Makefile
${CLUSTER_GEN}/remote-image: ${CLUSTER_GEN}/image-pushed .FORCE
```
If we remove the .FORCE, it might use a stale remote-image file if I'm not mistaken

kctf-operator/pkg/apis/kctf/v1alpha1/challenge_types.go
sroettger: I see, thanks for the clarification. Makes sense

base/Makefile
sirdarckcat: let's add it back :)

aliciafmachado: added back

aliciafmachado: i didn't know what it was for

sirdarckcat (approved):

PR #146, opened 2020-11-26
Fix apparmor and support updates

sroettger:
The apparmor profile had some spurious bytes at the beginning.
In addition, make the operator initialization update existing objects and bail out on any error.

### DIFF
--- a/kctf-operator/deploy/operator.yaml
+++ b/kctf-operator/deploy/operator.yaml
@@ -17,7 +17,7 @@ spec:
       containers:
         - name: kctf-operator
           # TODO: Replace this with the built image name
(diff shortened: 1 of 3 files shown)


sroettger:
so for every pull request, I need to wait for the builder and then manually tag the new version and update the yaml.
Wdyt, should we just change the yaml to use the ":beta" tag?

sroettger:
updated the operator version

sirdarckcat:
+1 on beta tag.

we should have a good workflow that doesn't require pushing, and then add the versions separately (maybe have a single commit, maybe off-head that explicitly sets the version, and then we can tag that?)

PR #147, opened 2020-11-26
Split docker workflows

sroettger:
Having one workflow per docker image to build allows us to have more fine grained control when to actually trigger the workflow.


sroettger:
I always merge with alpha at first by accident which triggers the sla bot :(. About time we make beta the default branch

sirdarckcat (approved):

PR #148, opened 2020-11-27
Unified workflow

sroettger:
Two changes:
* remove kctf-conf from samples directory (why did we have this checked in?)
* make the test workflow part of the build workflow

The new workflow has the following jobs:
* one job that builds all the docker images in parallel (using a job strategy)
* one job that
  * waits for the results
  * updates the new tags everywhere
  * builds/tests (the testing is blocked on working healthchecks)
  * push the changes if everything worked


sroettger:
The robot commit now happens with the email of the event head_commit.
And the test if it should be rebuilt is now based on the commit message instead of the email:
`git log -n1 --grep='Automated commit: update images.' --format=%H`

sirdarckcat (approved):

PR #153, opened 2020-12-04
Use full docker image instead of debootstrap for chroot

sroettger:


sirdarckcat (approved):
shpit!

sirdarckcat:
pls merge if the other tasks work =D

PR #156, opened 2020-12-07
annotate the domain properly

sirdarckcat:


sirdarckcat:
maybe we can do www.chal.domain for http, wdyt?

sroettger:
is this to distinguish between a tcp and http endpoint?
www. sounds fine too

sroettger (approved):

sirdarckcat:
yep, the http load balancers and tcp load balancers need different hosts. tasks could, in theory have both

sirdarckcat:
we can switch to www on beta

PR #161, opened 2020-12-09
Namespace overhaul

sroettger:
All challenges go into "default".
All kctf system stuff goes into "kctf-system".

This is much nicer to use. First of all, you see everything kctf related by listing the kctf-system namespace. E.g. not all kctf deployments have kctf in the name. It was not obvious to me that external-dns is part of kctf:
```sh
$kctf-kubectl get deployments --namespace kctf-system 
NAME            READY   UP-TO-DATE   AVAILABLE   AGE
external-dns    0/1     1            0           56m
kctf-operator   1/1     1            1           18m
```

And you can list all challenges, pods etc easily:
```sh
$kctf-kubectl get challenges                         
NAME         HEALTH    STATUS    DEPLOYED   PUBLIC
apache-php   healthy   Running   true       false
bash         healthy   Running   true       false
```


sroettger:
Deleting is super easy as well. The operator tracks which objects belong to the CRD and garbage collects them.

sroettger:
The makefile is getting more and more simple.
Remaining functionality is:
* local tests etc
* build the docker images and add them to the challenge.yaml
* kubectl apply / delete challenge.yaml

sroettger:
Finally the tests went through.

base/challenge-skeleton/challenge.yaml
```diff
@@ -2,7 +2,6 @@ apiVersion: kctf.dev/v1alpha1
 kind: Challenge
 metadata:
   name: mychal
-  namespace: mychal
```
sirdarckcat, line 5: samples

sirdarckcat:
does this affect the ability to do RBAC? IIRC namespaces are the level at which we are supposed to do multi-tenancy?

sroettger:
> does this affect the ability to do RBAC? IIRC namespaces are the level at which we are supposed to do multi-tenancy?

You should still be able to set the namespace if you want. It's just optional now.

sirdarckcat:
ok, I see. :) sounds good!

sirdarckcat (approved):

PR #165, opened 2020-12-11
Allow challenge to challenge communication

sroettger:
Adds a new field to challenge.Spec: "allowConnectTo" where you can specify other challenges you can connect to
* creates a service for every deployment
* if allowConnectTo is specified, create a network policy that allows egress to pods of those challenges
* plus a bugfix since we deleted persistentvolume claims by accident when moving everything to the default namespace

### DIFF
--- a/kctf-operator/deploy/crds/kctf.dev_challenges_crd.yaml
+++ b/kctf-operator/deploy/crds/kctf.dev_challenges_crd.yaml
@@ -133,6 +133,11 @@ spec:
                 items:
                   type: string
                 type: array
+              allowConnectTo:
+                description: List of challenge names that this challenge is allowed to connect to
+                items:
+                  type: string
+                type: array
               podTemplate:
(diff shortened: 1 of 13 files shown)


sirdarckcat:
if I understand correctly, the reason we do the ACL on the sender rather than receiver, is because the only thing stopping a task form talking to another task is the network policy forbidding it, so we revoke that restriction narrowly.

I think it's an odd choice, so it might make sense to explain it somewhere, although it makes sense given the way things work.

sirdarckcat:
oh, and also might be worth mentioning this isn't compatible with HTTPS services, because internal ingress wouldn't work? although I don't expect this to be a problem, as most HTTPS services also have a LoadBalancer IIRC

sirdarckcat (approved):

sirdarckcat:
we can add that in docs later, though.. no need to be part of this change

PR #168, opened 2020-12-16
Run the proof-of-work outside of the chroot

sroettger:
Currently, we add proof-of-work related files to the player's chroot.
This adds requirements to the chroot and adds visible infrastructre details to the challenge environment.

We can avoid this by running the proof-of-work outside of the chroot and use socat instead of nsjail's tcp mode.


sroettger:
Note that I only updated the bash sample so far.

sroettger:
I believe I updated all challenges now, let's see if the tests pass.

samples/bash/challenge/image/Dockerfile
```diff
@@ -18,7 +18,6 @@ COPY nsjail.cfg /home/user/
 
 CMD kctf_setup && \
     drop_privs \
-    nsjail --config /kctf/nsjail_base.cfg --config /home/user/nsjail.cfg -- \
-    /usr/bin/stdbuf -i0 -o0 -e0 \
-    maybe_pow.sh \
-    /bin/bash
+    socat \
+      TCP-LISTEN:1337,reuseaddr,fork \
+      EXEC:"/usr/bin/stdbuf -i0 -o0 -e0 maybe_pow.sh nsjail --config /home/user/nsjail.cfg -- /bin/bash"
```
sirdarckcat: Can we make this a shell script? That just accepts params

sirdarckcat (approved):
One suggestion below

samples/bash/challenge/image/Dockerfile
sroettger: And ship it as part of the nsjail image? (we should btw rename that image, kctf-base or something)

I'm torn. I think it would look nicer but it also hides the details how this works and it might prevent challenge authors to make changes (maybe the script will do other required stuff that they don't know or start doing it in the future).

sroettger:
It looks like the stdbuf might still be needed. Need to add it back.

sroettger:
Actually, should we have stdbuf in there? It's not needed for the sample challenges so it feels wrong to set it in general.

The pro for having it is: it's a common issue when writing challenges and this could save some headache.
The con: unnecessary for most challenges, you can call flush instead in your binary when needed.

I'm leaning to not having it and leave it to challenge authors if it's needed.

sroettger:
rebased against the other pull requests

sirdarckcat:
What about a nsjail_pow.sh that just summarizes stdbuf, maybepow and nsjail? (But we keep socat and pass all arguments down to nsjail).

Like..
```
socat tcp-l:1337,forever system:'/kctf/pow_nsjail --config /kctf/nsjail.conf -- /bin/bash'
```

Still ugly but more readable

samples/bash/challenge/image/Dockerfile
sirdarckcat: Sent another suggestion below. Not blocking though.

sroettger:
I tested this but it looks like the stdbuf needs to come after nsjail. So I think we should just not do stdbuf ourselves.

sirdarckcat:
sg, I merged your other commit so now this one has conflicts

PR #169, opened 2020-12-16
Run as uid 1000 by default with ProcMount: Unmasked

sroettger:
The procmount: unmasked features seems to be available now.
This allows us to run as an unprivileged user by default.

Apache-php and apache-other are still root. We could probably run them as the user if we change the port or give cap_net_smth, but maybe it's ok like this.
The xss-bot needs to mount a tmpfs at /tmp. I replaced this with an EmptyDir volume.

Open questions:
* should a challenge always have a working /tmp? Or maybe we could expose an option for that since changing the PodTemplate is not very beautiful.
* should we expoose "runAsRoot" as an option in the challenge spec?


sirdarckcat (approved):
Woohoo!

Apache could run in a different port, but yea sgtm for now.

PR #182, opened 2021-01-08
Allow challenges to be in any subdirectory

### LINKED ISSUE #142: Support custom directory layouts
A common directory layout is to have a folder for every challenge category:
* pwn/foo
* pwn/bar
* web/baz

sroettger:
Some CTFs want to sort their challenges by category, support arbitrary subdirectories for challenges.
I.e. you can now do:
```
myctf/pwn/foo
myctf/pwn/bar
myctf/crypto/baz
```
etc.

To make it clear that you can do this, I added a `--challenge-dir` flag to the `kctf chal create` command.

Fixes: #142

### DIFF
--- a/dist/bin/kctf-challenge
+++ b/dist/bin/kctf-challenge
@@ -403,6 +403,15 @@ function kctf_chal_create {
 
   if [[ -z "${CHALLENGE_DIR}" ]]; then
     CHALLENGE_DIR="${KCTF_CTF_DIR}/${CHALLENGE_NAME}"
+  else
+    CHALLENGE_DIR_REALPATH=$(realpath --canonicalize-missing "${CHALLENGE_DIR}")
+    if [[ "${CHALLENGE_DIR_REALPATH}" != "${KCTF_CTF_DIR}"/* ]]; then
+      echo "Challenge dir needs to be under the CTF dir:" >&2
+      echo "  \"${CHALLENGE_DIR_REALPATH}\"" >&2
+      echo "  not under" >&2
+      echo "  \"${KCTF_CTF_DIR}\"" >&2
+      exit 1
+    fi
   fi
   if [[ -e "${CHALLENGE_DIR}" ]]; then
     echo "challenge dir \"${CHALLENGE_DIR}\" does already exist" >&2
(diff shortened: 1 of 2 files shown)


dist/bin/kctf-challenge
```diff
@@ -352,47 +352,92 @@ function kctf_chal_debug {
...
   shift
 
-  #we don't require anymore that the path is under the ctf directory.
-  #TODO: add a flag to control this
-  CHALLENGE_DIR="${KCTF_CTF_DIR}/${CHALLENGE_NAME}"
+  if [[ -z "${CHALLENGE_DIR}" ]]; then
+    CHALLENGE_DIR="${KCTF_CTF_DIR}/${CHALLENGE_NAME}"
+  else
+    CHALLENGE_DIR_REALPATH=$(realpath --canonicalize-missing "${CHALLENGE_DIR}")
+    if [[ "${CHALLENGE_DIR_REALPATH}" != "${KCTF_CTF_DIR}"/* ]]; then
```
sirdarckcat, line 408: why is this necessary?

dist/bin/kctf-challenge
```diff
@@ -352,47 +352,92 @@ function kctf_chal_debug {
...
+      echo "Challenge dir needs to be under the CTF dir:" >&2
+      echo "  \"${CHALLENGE_DIR_REALPATH}\"" >&2
+      echo "  not under" >&2
+      echo "  \"${KCTF_CTF_DIR}\"" >&2
+      exit 1
+    fi
+  fi
   if [[ -e "${CHALLENGE_DIR}" ]]; then
+    echo "challenge dir \"${CHALLENGE_DIR}\" does already exist" >&2
```
sirdarckcat: > challenge dir ... already exists

using does here sounds more like "challenge dir indeed exists already"

sirdarckcat (approved):

dist/bin/kctf-challenge
sroettger: we want to rely on this for versioning. I.e. we want to rely on the fact that every challenge is bound to the kctf version.
That's why I think it's good to (soft) enforce that challenges need to be under the CTF directory.

dist/bin/kctf-challenge
sroettger: added "error: " in front

dist/bin/kctf-challenge
sirdarckcat: k

PR #183, opened 2021-01-08
Service/Ingress changes

sroettger:
A few changes related to services/ingresses:
* Use a NodePort service as the IngressBackend
* The NodePort service replaces the internal ClusterIP service
* Change ports to default 1337 and default port == targetPort
* Delete challenges before deleting the cluster to make sure that load balancers get cleaned up

### DIFF
--- a/dist/bin/kctf-cluster
+++ b/dist/bin/kctf-cluster
@@ -125,6 +125,13 @@ function kctf_cluster_start {
 }
 
 function kctf_cluster_stop {
+  echo "deleting all challenges so that load balancers etc can be cleaned up"
+  for chal_name in $(kubectl get challenges -o=jsonpath="{.items[*].metadata.name}"); do
+    kubectl delete "challenge/${chal_name}"
+  done
+  # deleting the cluster below takes a while, so sleeping for a bit doesn't hurt
+  sleep 20
+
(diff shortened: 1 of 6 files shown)


dist/bin/kctf-cluster
```diff
@@ -125,6 +125,13 @@ function kctf_cluster_start {
 }
 
 function kctf_cluster_stop {
+  echo "deleting all challenges so that load balancers etc can be cleaned up"
```
sirdarckcat, line 128: can we delete the default namespace?

dist/bin/kctf-cluster
```diff
@@ -125,6 +125,13 @@ function kctf_cluster_start {
...
+  for chal_name in $(kubectl get challenges -o=jsonpath="{.items[*].metadata.name}"); do
+    kubectl delete "challenge/${chal_name}"
+  done
+  # deleting the cluster below takes a while, so sleeping for a bit doesn't hurt
+  sleep 20
```
sirdarckcat, line 135: maybe echo "Waiting ... " or something?

sirdarckcat (approved):

dist/bin/kctf-cluster
sroettger: I don't think so.
That reminds me, this will fail if someone runs challenges in a different namespace. We should probably search for challenges in all namespaces.

sirdarckcat: ah, yea

PR #184, opened 2021-01-11
Add support for "kind" clusters back

sroettger:
You can now create a --type "kind" cluster with `kctf config create` and then use it as usual.
I.e. just run `kctf chal start` and it will deploy to the local cluster.

The apparmor handling is copied from how it worked before, but it's racy since the deployment might not have been created yet.
I'm thinking of exposing an apparmor option in the challenge.yaml to control this.
Wdyt?

### DIFF
--- a/dist/activate
+++ b/dist/activate
@@ -15,6 +15,8 @@
 KCTF_YQ_URL="https://github.com/mikefarah/yq/releases/download/v4.2.0/yq_linux_amd64"
 KCTF_YQ_HASH="5d44bd64e264e9029c5f06bcd960ba162d7ed7ddd1781f02a28d62f50577b632"
 
+KIND_URL="https://kind.sigs.k8s.io/dl/v0.9.0/kind-$(uname)-amd64"
+
 export KCTF_CTF_DIR="$(realpath --no-symlinks "$(dirname "${BASH_SOURCE-$0}")/..")"
 
 function _kctf_setup_environment {
@@ -38,6 +40,11 @@ function _kctf_download_dependencies {
     wget "${KCTF_YQ_URL}" -O "${KCTF_BIN}/yq" --quiet || return 1
     sha256sum --status -c <(echo "${KCTF_YQ_HASH}  ${KCTF_BIN}/yq") || return 1
     chmod u+x "${KCTF_BIN}/yq"
+  fi
+
+  if [[ ! -x "${KCTF_BIN}/kind" ]]; then
+    curl -Lo "${KCTF_BIN}/kind" "${KIND_URL}" || return 1
+    chmod u+x "${KCTF_BIN}/kind"
   fi
 }
 
--- a/dist/activate
+++ b/dist/activate
@@ -104,6 +104,7 @@ function kctf {
           return 1
         fi
         source "${KCTF_CTF_DIR}/kctf/config/${CONFIG_NAME}"
+        export CLUSTER_TYPE
         export PROJECT
         export ZONE
         export REGISTRY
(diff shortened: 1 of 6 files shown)


dist/activate
```diff
@@ -15,6 +15,8 @@
 KCTF_YQ_URL="https://github.com/mikefarah/yq/releases/download/v4.2.0/yq_linux_amd64"
 KCTF_YQ_HASH="5d44bd64e264e9029c5f06bcd960ba162d7ed7ddd1781f02a28d62f50577b632"
 
+KIND_URL="https://kind.sigs.k8s.io/dl/v0.9.0/kind-$(uname)-amd64"
```
sirdarckcat: shoul we do the hash thing too?

sirdarckcat: Also, we can probably hardcode Linux (like with yq)

dist/bin/kctf-cluster
```diff
@@ -14,7 +14,27 @@
...
+
+  # The operator needs to create some subresources, e.g. the gcsfuse service account
+  for i in {1..30}; do
+    kubectl get pods --namespace kctf-system -o=jsonpath="{.items[*].status.containerStatuses[?(@.imageID==\"${OPERATOR_IMAGE}\")].ready}" | grep "true" && break
+    if [ "$i" == "30" ]; then
+      echo "Couldn't find a kctf-operator pod with status ready=true and image=\"${OPERATOR_IMAGE}\" after 5 minutes" >&2
+      kubectl get pods --namespace kctf-system -o=yaml >&2
+      exit 1
+    fi
+    sleep 10
```
sirdarckcat: nit:Maybe do 3 seconds (and 100x loop)? It's otherwise a bit boring to see no output.

dist/bin/kctf-config
```diff
@@ -176,11 +196,24 @@ function kctf_config_create {
...
+    gce)
+      if [[ -z "$PROJECT" ]]; then
+        echo "Missing required argument \"--project\"." >&2
+        kctf_config_create_usage
+        return 1
+      fi
+      ;;
+    kind)
+      if [[ -z "$PROJECT" ]]; then
+        PROJECT="kind-project"
```
sirdarckcat: I worry a bit about the project ending up somewhere that does a fetch (eg, gcr) can we use a project we own? Just in case.

sirdarckcat (approved):

sirdarckcat:
Apparmor is a cluster thing, not a challenge thing, isn't it? Why would it be in challenge?

sroettger:
The challenge needs to specify which apparmor profile it should run with.
This breaks because we don't have apparmor in kind.
We could also decide to remove the apparmor profile, not sure how much security it provides and if it's worth the complexity.

sirdarckcat:
I remember someone said that apparmor didn't help us because we allow mounts, but lets check

dist/bin/kctf-config
sroettger: We could also change the code to use a different image URL, I think that's the only reason why I had to set this.

dist/activate
sroettger: yeah, done.

dist/bin/kctf-cluster
sroettger: done

dist/bin/kctf-config
sroettger: done, it's not using the PROJECT anymore

sroettger:
I think the apparmor change should be a separate PR since it's quite an important change.
So will submit this if the checks pass.

PR #185, opened 2021-01-12
Fixes to support multiple namespaces

sroettger:
We don't specify the namespace in various commands. Fix that by reading it with yq and then using the namespace everywhere we run kubectl.


dist/activate
```diff
@@ -153,12 +153,17 @@ function _kctf_set_active_challenge {
   current_dir="${PWD}"
   while [[ "${current_dir}" == "${KCTF_CTF_DIR}"/* ]]; do
     if [[ -e "${current_dir}/challenge.yaml" ]]; then
-      CHALLENGE_NAME=$("${KCTF_BIN}/yq" eval '.metadata.name' "${current_dir}/challenge.yaml" 2>/dev/null)
+      CHALLENGE_NAME=$("${KCTF_BIN}/yq" eval --exit-status '.metadata.name' "${current_dir}/challenge.yaml" 2>/dev/null)
       if [[ $? -ne 0 ]]; then
         unset CHALLENGE_NAME
       fi
+      CHALLENGE_NAMESPACE=$("${KCTF_BIN}/yq" eval --exit-status '.metadata.name' "${current_dir}/challenge.yaml" 2>/dev/null)
```
sirdarckcat: .namespace?

sirdarckcat (changes requested):

dist/activate
sroettger: done

sirdarckcat (approved):

PR #189, opened 2021-01-12
Nicer log output

sroettger:
The kctf functions will now log with 
[*]
[W]
[E]
for info, warn and error respectively. These prefixes are green / yellow and red.


sirdarckcat (approved):
haha showoff! :P

Issue #110, opened 2020-06-30
Make it possible to provision SSL certificates before CTF starts


sirdarckcat:
According to https://cloud.google.com/load-balancing/docs/ssl-certificates/troubleshooting#certificate-managed-status  Provisioning a Google-managed certificate might take up to 60 minutes. This makes it unusable during a CTF in practice.

Note that a possibly worse problem is leaking challenge names before the CTF starts - see for example https://crt.sh/?q=kctf.dev

sirdarckcat:
this possibly could be done using a wildcard SSL certificate, which Google doesn't provide, but LetsEncrypt does. We just need DNS changes https://github.com/ahmetb/gke-letsencrypt

sirdarckcat:
https://certbot-dns-google.readthedocs.io/en/stable/

sirdarckcat:
https://github.com/acmesh-official/acme.sh/wiki/dnsapi#49-use-google-cloud-dns-api-to-automatically-issue-cert

sirdarckcat:
note sometimes the certificate is issued within 10 minutes =/

sirdarckcat:
we can do this somewhat easily if we use letsencrypt with wildcard certs and a single ingress.

the problem of using a single ingress, though, is that then we might make some tasks easier/harder, as we are adding a proxy (GFE)

sirdarckcat:
this: https://cert-manager.io/docs/configuration/acme/dns01/google/#gke-workload-identity

seems very straightforward, but adds a dependency for kctf on this project

sirdarckcat:
another possible solution is to point the ingress to an empty default backend, so the certificate is issued before the challenge is public.

sirdarckcat:
that means that here we dont delete

https://github.com/google/kctf/blob/5ce3c754741d2b5a8693c6ebf537d0d2786e69f2/kctf-operator/pkg/controller/challenge/service/functions.go#L108

and instead here

https://github.com/google/kctf/blob/5ce3c754741d2b5a8693c6ebf537d0d2786e69f2/kctf-operator/pkg/controller/challenge/service/service.go#L89

we select another service/port (we have to generate a fake service/port) if challenge is not public

sirdarckcat:
hmm we apparently can "just" point to an ExternalName, possibly even make it point to a domain that doesn't exist

sirdarckcat:
Looks like we can use an ExternalName pointing to 
```
task-name.kctf.invalid
```

[...]

PR #200, opened 2021-01-14
Add a letsencrypt/ACME client

sirdarckcat:
The address of the docker image will be changed in a follow-up PR.

Ref #110

### DIFF
--- a/dist/bin/kctf-config
+++ b/dist/bin/kctf-config
@@ -113,6 +113,9 @@ function kctf_config_create_usage {
   echo -e "                  Possible values are us.gcr.io, asia.gcr.io, and eu.gcr.io" >&2
   echo -e "  --cluster-name  Name of the kubernetes cluster (default: kctf-cluster)" >&2
   echo -e "  --domain-name   Optional domain name to host challenges under" >&2
+  echo -e "  --email-address Optional email address for LetsEncrypt registration (for wildcard certificates)" >&2
+  echo -e "                  To use it, please read and agree to the ACME Subscriber Agreement:" >&2
+  echo -e "                    https://letsencrypt.org/repository/" >&2
   echo -e "  --start-cluster Start the cluster if it's not running yet" >&2
 }
 
@@ -124,10 +127,11 @@ function kctf_config_create {
   ZONE="europe-west4-b"
   CLUSTER_NAME="kctf-cluster"
   DOMAIN_NAME=""
+  EMAIL_ADDRESS=""
   START_CLUSTER="0"
 
   OPTS="h"
-  LONGOPTS="help,type:,project:,zone:,registry:,cluster-name:,domain-name:,start-cluster"
+  LONGOPTS="help,type:,project:,zone:,registry:,cluster-name:,domain-name:,email-address:,start-cluster"
   PARSED=$(getopt --options=$OPTS --longoptions=$LONGOPTS --name "kctf config create" -- "$@")
   if [[ $? -ne 0 ]]; then
     kctf_config_create_usage
@@ -163,6 +167,10 @@ function kctf_config_create {
         ;;
       --domain-name)
         DOMAIN_NAME="$2"
+        shift 2
+        ;;
+      --email-address)
+        EMAIL_ADDRESS="$2"
         shift 2
         ;;
       --start-cluster)
@@ -229,6 +237,7 @@ ZONE=${ZONE}
 REGISTRY=${REGISTRY}
 CLUSTER_NAME=${CLUSTER_NAME}
 DOMAIN_NAME=${DOMAIN_NAME}
+EMAIL_ADDRESS=${EMAIL_ADDRESS}
 EOF
 
   case "${CLUSTER_TYPE}" in
@@ -267,6 +276,7 @@ EOF
     export REGISTRY
     export CLUSTER_NAME
     export DOMAIN_NAME
+    export EMAIL_ADDRESS
     "${KCTF_BIN}/kctf-cluster" start >&2 || return 1
   fi
 
--- a/docker-images/certbot/certbot.sh
+++ b/docker-images/certbot/certbot.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+
+TEST="--test-cert"
+
+if [ -z "${DOMAIN}" ]; then
+  echo Make sure the DOMAIN environment variable points to the domain.
+  exit 1
+fi
+
+if [ -z "${EMAIL}" ]; then
+  echo Make sure the EMAIL environment variable points to the email.
+  exit 1
+fi
+
+if [ -z "${SECRET}" ]; then
+  echo Make sure the SECRET environment variable points to the k8s secret.
+  exit 1
+fi
+
+if [ -z "${PROD}" ]; then
+  echo Making a test certificate because PROD environment variable is not set.
+else
+  echo Making a valid certificate because PROD environment variable is set.
+  TEST=""
+fi
+
+while true; do
+  echo Waiting 2 minutes to avoid hitting rate limits
+  sleep 2m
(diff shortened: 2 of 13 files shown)


kctf-operator/pkg/resources/external-dns.go
```diff
@@ -67,24 +72,58 @@ func NewExternalDnsDeployment() runtime.Object {
...
+										Key: "DOMAIN_NAME",
+									},
+								},
+							}},
+							Args: []string{"--log-level=debug", "--source=service", "--source=ingress",
+								"--provider=google", "--domain-filter=$(DOMAIN_NAME)", "--registry=txt",
+								"--txt-owner-id=kctf-cloud-dns"},
+						},
+						{
+							Image: "eu.gcr.io/sdcpocs1/certbot",
```
sirdarckcat, line 95: this will be changed in the next PR, after I add the docker-image to the workflow

sroettger: why not update the workflow as part of this PR?

dist/bin/kctf-config
```diff
@@ -113,6 +113,9 @@ function kctf_config_create_usage {
   echo -e "                  Possible values are us.gcr.io, asia.gcr.io, and eu.gcr.io" >&2
   echo -e "  --cluster-name  Name of the kubernetes cluster (default: kctf-cluster)" >&2
   echo -e "  --domain-name   Optional domain name to host challenges under" >&2
+  echo -e "  --email-address Optional email address for LetsEncrypt registration (for wildcard certificates)" >&2
+  echo -e "                  To use it, please read and agree to the ACME Subscriber Agreement:" >&2
```
sroettger, line 117: what happens if you don't set the email address?

docker-images/certbot/certbot.sh
```diff
@@ -0,0 +1,38 @@
...
+while true; do
+  echo Waiting 2 minutes to avoid hitting rate limits
+  sleep 2m
+  if ./kubectl get secret "${SECRET}" -o 'jsonpath={.data}' | jq -r '.["tls.crt"]' | base64 -d | openssl x509 -checkend 2592000 -noout -in -; then
+    echo Certificate is valid for at least 30 days
+    sleep 15d
+  else
+    certbot certonly "${TEST}" --non-interactive --agree-tos -m "${EMAIL}" --dns-google -d '*.'"${DOMAIN}" --dns-google-propagation-seconds 120 && \
+    (./kubectl create secret tls "${SECRET}" --cert /etc/letsencrypt/live/"${DOMAIN}"/fullchain.pem --key /etc/letsencrypt/live/"${DOMAIN}"/privkey.pem --dry-run=client --save-config -o yaml | ./kubectl apply -f -) && \
+    echo Created and saved certificate
```
sroettger: Should this also update the cert in all other namespaces?

docker-images/certbot/certbot.sh
```diff
@@ -0,0 +1,38 @@
...
+
+while true; do
+  echo Waiting 2 minutes to avoid hitting rate limits
+  sleep 2m
+  if ./kubectl get secret "${SECRET}" -o 'jsonpath={.data}' | jq -r '.["tls.crt"]' | base64 -d | openssl x509 -checkend 2592000 -noout -in -; then
+    echo Certificate is valid for at least 30 days
+    sleep 15d
+  else
+    certbot certonly "${TEST}" --non-interactive --agree-tos -m "${EMAIL}" --dns-google -d '*.'"${DOMAIN}" --dns-google-propagation-seconds 120 && \
+    (./kubectl create secret tls "${SECRET}" --cert /etc/letsencrypt/live/"${DOMAIN}"/fullchain.pem --key /etc/letsencrypt/live/"${DOMAIN}"/privkey.pem --dry-run=client --save-config -o yaml | ./kubectl apply -f -) && \
```
sroettger: maybe this could be in a function so that it's easier to read?
and does it need an else block to print an error or do you think this is clear from the kubectl output

dist/bin/kctf-config
sirdarckcat: You don't get a wildcard certificate, only a ManagedCert

docker-images/certbot/certbot.sh
sirdarckcat: This is done by the operator below

dist/bin/kctf-config
sirdarckcat: This might be interesting to clarify - the LetsEncrypt certificate only helps with reducing the latency, but you still are bound to DNS delays and LoadBalancer delays.

The DNS delays now are the slowest part of the process, but fixing that isn't easily possible unless we reserve IP addresses and assign them to LoadBalancers ahead of time. Reserving an IP costs money, and I'm not sure the extra 1-2 minutes are worth it.

We could make it an option in a future version.

docker-images/certbot/certbot.sh
sirdarckcat: To clarify - the operator listens on changes to Challenge. As soon as any challenge changes (or gets uploaded), the certificate will be propagated to that challenge namespace.

There's a slight race condition, if, for example, you create a challenge and make it public immediately, then the Ingress might be created before the LetsEncrypt certificate is available. If that ever happens, the ManagedCertiticate will be used instead.

kctf-operator/pkg/resources/external-dns.go
sirdarckcat: I want to experiment with it a bit. Maybe I can extract this information from config, rather than sed-ing the go code.

I don't really know of a good way to play with workflows short of just commiting things directly to github :-P

The issuer is currently not enabled (the PROD env variable is not set, so the certificates aren't being issued), if you were worried about that (also, we probably wanna keep it that way for the workflow tests)

docker-images/certbot/certbot.sh
sroettger: What if no challenge changes?
I.e. I host challenges for a long time without touching them and the certbot updates the certificate. The operator would never notice this, no?

sirdarckcat: yea that's a problem.

https://github.com/google/kctf/pull/200/commits/565b66bb651a3bfa325c1d3bdb9a3f021fc3525f#diff-c8b71c089af554bf4958c42612d1cb8421169936d5b4d18355b595876374d4ac

docker-images/certbot/certbot.sh
sirdarckcat: how would the function look like?

```bash
update_cert() {
   certbot ...
   if ! $?; then
        ...
   else
       ...
   fi
   ....
}
```

or did you have something else in mind

sroettger: I was thinking something like:
```sh
function update_cert() {
  certbot certonly "${TEST}" --non-interactive --agree-tos -m "${EMAIL}" --dns-google -d '*.'"${DOMAIN}" --dns-google-propagation-seconds 120 || return
  ./kubectl create secret tls "${SECRET}" --cert /etc/letsencrypt/live/"${DOMAIN}"/fullchain.pem --key /etc/letsencrypt/live/"${DOMAIN}"/privkey.pem --dry-run=client --save-config -o yaml | ./kubectl apply -f -
}

if update_cert; then
  echo 'success'
else
  echo 'fail'
fi
```

sroettger (approved):

docker-images/certbot/certbot.sh
sirdarckcat: ok, done

kctf-operator/pkg/resources/external-dns.go
sirdarckcat: I'll also fix GCSFUSE as that is doing `apt install` at bootstrap, which seems suboptimal

PR #222, opened 2021-02-15
Adding mention of SSH/VPN to tunnel traffic.

zaeyx:
Small change here, want to mention that a trusted server can be used as a 'proxy' (jumpbox) for accessing the challenge.  Having players tunnel directly to that trusted server can help avoid MiTM style interception of their connections.

### DIFF
--- a/docs/security-threat-model.md
+++ b/docs/security-threat-model.md
@@ -25,6 +25,7 @@ As such, the risks in this area are:
 
 Ways to limit these risks are:
  - Players could connect to the tasks through HTTPS. However, there are currently no plans to implement this functionality.
+ - Players may 'tunnel' via VPN/SSH to a protected proxy device from which they then access the CTF.
 
 The compromise of the players themselves is out of scope of kCTF's threat model.
 


sirdarckcat:
Hi

Im not 100% sure what this is meant to convey. Would you mind elaborating a bit?

zaeyx:
Hey, absolutely.  Here's what I'm trying to convey (though in as few words as possible):

A mechanism for avoiding the interception of traffic MiTM style between CTF participants and any service endpoints is - to have participants connect from an untrusted network (perhaps their home?) to the trusted network via a 'jumpbox'.  A simple way to do this would be to have a trusted network from where the CTF is run, then have participants create an ssh session with a device on that network (that has a publicly accessible SSH server); and tunnel their traffic through that SSH server.

So now if there was someone trying to intercept participant's communications - that someone would have to be on the trusted network, between the 'jumpbox' (SSH server) and the CTF components.  

I'm happy to rephrase of course.  Does my explanation here make more sense?  Should I re-write the proposed addition to try and clarify further?

sirdarckcat (approved):

PR #231, opened 2021-02-25
Don't automatically try to upgrade the cluster on config create

sroettger:
We shouldn't touch an existing cluster unexpectedly.
This was nearly an issue for me when I wanted to upgrade an existing cluster. I.e. I wanted to load the config without changing the cluster, which was not possible.
Now, just warn the user if there's an existing cluster.

### DIFF
--- a/.gitallowed
+++ b/.gitallowed
@@ -1,0 +1,1 @@
+KCTF_CLOUD_API_KEY.*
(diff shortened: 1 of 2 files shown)


sirdarckcat (approved):

sirdarckcat:
```
/tmp/samples/kctf/bin/kctf-cluster: line 289: syntax error near unexpected token `;'
/tmp/samples/kctf/bin/kctf-cluster: line 289: `  if [[ "$***START_CLUSTER***" == "1" ]] || ; then'
```

PR #251, opened 2021-03-05
connect to most recent running running pod with ssh

### LINKED ISSUE #235: Debug challenge that starts failing after changes
If you make a change that breaks a challenge and run chal start, you can't debug it since the old version is still running.

We could have an option to choose which pod to debug, or always attach to the newest pod or so.

sroettger:
Fixes #235

### DIFF
--- a/dist/bin/kctf-challenge
+++ b/dist/bin/kctf-challenge
@@ -241,7 +241,35 @@ function kctf_chal_debug_ssh {
   require_cluster_config
   COMMAND="debug ssh" parse_container_name $@ || return
 
-  "${KCTF_BIN}/kubectl" exec "deployment/${CHALLENGE_NAME}" --namespace "${CHALLENGE_NAMESPACE}" -c "${CONTAINER}" -it -- /bin/bash
+  pods=($("${KCTF_BIN}/kubectl" get pods -o jsonpath='{.items[*].metadata.name}'))
+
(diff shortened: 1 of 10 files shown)


dist/bin/kctf-challenge
```diff
@@ -241,7 +241,35 @@ function kctf_chal_debug_ssh {
...
+
+  if [[ ${#pods[@]} -eq 0 ]]; then
+    _kctf_log_err 'No pods found. Is the challenge running?'
+    return 1
+  fi
+
+  pod="${pods[0]}"
+  if [[ ${#pods[@]} -ne 1 ]]; then
+    _kctf_log "Found ${#pods[@]} pods, connecting to the most recent one."
+    _kctf_log "To connect to another pod, run 'kubectl exec pod/PODNAME --namespace ${CHALLENGE_NAMESPACE} -c ${CONTAINER} -it -- /bin/bash'"
```
sirdarckcat: you can say `deployment/${challenge_name}` to connect to the "running" one, and that way users dont need to understand which pods there are.

or you can print the other pod names.

sirdarckcat (approved):

dist/bin/kctf-challenge
sroettger: I thought about this too but decided not to add it since the list of pod names is not really useful to choose from.
I added a note now how to list the pods.

PR #253, opened 2021-03-08
remove chars that showed up in the bash prompt

sroettger:


sroettger:
the prompt is still broken in bash though if you type more than one line :(

sirdarckcat (approved):

sirdarckcat:
Well I wish I understood this..

Issue #246, opened 2021-03-05
Do an end-to-end test with a new project


sroettger:
Make sure that the chal create doesn't block on any errors/questions.
Also that all APIs are enabled etc.

sirdarckcat:
I tested on kind, and found some issues I left on #228 for docs.
We should test with gcloud too, but those probably can also be fixed with docs for #228.

sroettger:
Working on the end to end test from a new project now.
Some notes on the steps:
* created an ubuntu 20.04 vm in GCE and deleted the google-cloud-sdk snap that came with it
* Opened the docs at https://google.github.io/kctf/ and started with local testing
* docker install, works
* downloading kctf etc instructions outdated as expected
* mkdir e2e, curl kctf | tar && cd e2e
* source kctf/activate
* kctf cluster create --type kind --start local
* kctf chal create --template pwn chal1 && cd chal1
* kctf chal start
* kctf chal status
* kctf chal debug port-forward
* nc 0 PORT
* kctf cluster stop

Then on GCP:
* create a project in pantheon
* try to run kctf cluster create --project <tab> => prints "gcloud not installed in autocomplete"
* kctf cluster create --project tmp-12317236178263 --domain-name tmp-12317236178263.kctf.cloud gcptest
  * prints warning "configuration created only locally, gcloud not installed"
* sudo snap install --classic google-cloud-sdk (--classic means it uses no sandboxing)
* kctf cluster start
  * ERROR: (gcloud.compute.networks.list) Some requests did not succeed:
     - Insufficient Permission: Request had insufficient authentication scopes.
* gcloud auth login
  * ERROR: (gcloud.auth.login) Cannot set property in configuration [kctf-6fb733c3c45c462b35c7], it does not exist.
* deactivate
* source kctf/activate
* gcloud auth login
  * You are now logged in
* kctf cluster start
  * [E] No cluster config loaded
* confusing: $(_kctf_config_string) returns ,config=gcptest
  * need to unset KCTF_CONFIG (also CHALLENGE_NAMESPACE, CHALLENGE_NAME, CHALLENGE_DIR) on deactivate
* loading .lastconfig fails silently => problem: required gcloud container cluster get credentials
* kctf cluster start
  * - Insufficient Permission: Request had insufficient authentication scopes.
* deactivate, gcloud auth login, source kctf/activate
  * blocked until I pressed enter, no idea what it was blocked on
* kctf cluster start
[...]

PR #255, opened 2021-03-08
various fixes from e2e test

sroettger:
This addresses the issues I described in #246.

### DIFF
--- a/dist/bin/kctf-cluster
+++ b/dist/bin/kctf-cluster
@@ -23,6 +23,7 @@ KCTF_CLOUD_API_KEY="AIzaSyC7Jgu4e0IygmImZNPmJHrcfZ3lJA9ZrZs"
 function update_gcloud_config {
   _kctf_log "Updating gcloud config."
   ACTIVE_ACCOUNT="$(CLOUDSDK_ACTIVE_CONFIG_NAME= gcloud config get-value core/account 2>/dev/null)"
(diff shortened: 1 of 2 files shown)


sroettger:
It probably makes sense to look at the commits one by one since they are mostly unrelated. It just felt like lots of overhead to create a PR for each of those. But let me know if you prefer me to split them.

dist/bin/kctf-cluster
```diff
@@ -263,8 +265,19 @@ EMAIL_ADDRESS=${EMAIL_ADDRESS}
...
+  set_lastconfig_link || return
+
   case "${CLUSTER_TYPE}" in
     gce)
+      if ! command -v gcloud >/dev/null 2>&1; then
+        if [[ "${START_CLUSTER}" == "1" ]]; then
+          _kctf_log_err "Can't start cluster, configuration created only locally. Gcloud not installed."
+          return 1
+        else
+          _kctf_log_warn "Configuration created only locally. Gcloud not installed."
```
sirdarckcat, line 277: Is the GCloud relevant to the user? Should we warn when starting instead only?

sirdarckcat (approved):

dist/bin/kctf-cluster
sroettger: I believe it's something we should warn about. It's a weird case if you create a gce config without having gcloud installed.

PR #261, opened 2021-03-13
Update docs

### LINKED ISSUE #228: Write new docs
1. Update cloudshell demo to use sdk
2. Update cloudshell demo to use kind
3. Add new step to cloudshell demo for gke
4. Make a 3 step quick start for landing
5. Update all other docs (threat model, etc)

sirdarckcat:
Fixes #228


docs/google-cloud.md
```diff
@@ -21,9 +21,7 @@ Set up billing, and enable the compute API:
 
 You can enable APIs from the command line with:
 ```
-gcloud services enable compute container containerregistry.googleapis.com
-gcloud services enable compute container compute.googleapis.com
-gcloud services enable compute container container.googleapis.com
```
sroettger, line 26: why are these not needed anymore?

docs/google-cloud.md
```diff
@@ -21,9 +21,7 @@ Set up billing, and enable the compute API:
...
+gcloud services enable compute container containerregistry.googleapis.com dns
```
sroettger: you probably want to do that after kctf/activate && the config has been created

docs/google-cloud.md
```diff
@@ -42,18 +40,18 @@ umask 0022
...
 ```
 
-## Create the cluster
-
-### Run the configuration script
-
+### Create the GKE cluster
 ```
-kctf-config-create --chal-dir ~/kctf-demo --project {{project-id}} --start-cluster
+kctf config create --project {{project-id}} --domain-name {{project-id}}-codelab.kctf.cloud --start remote-cluster
```
sroettger, line 60: maybe add a note that the user should assume that the domain name + challenge names will be public

docs/google-cloud.md
```diff
@@ -92,44 +90,43 @@ In this step, you'll learn how to:
 ## Call kctf-chal-create.sh to copy the skeleton
```
sroettger: this is outdated

docs/google-cloud.md
```diff
@@ -92,44 +90,43 @@ In this step, you'll learn how to:
...
 Run the following command to create a challenge called "demo-challenge":
 ```
-kctf-chal-create demo-challenge
+kctf chal create demo-challenge && cd demo-challenge
```
sroettger, line 101: should we introduce the template parameter here?

docs/google-cloud.md
```diff
@@ -92,44 +90,43 @@ In this step, you'll learn how to:
...
 
 This command deploys the image to your cluster, which will soon start consuming CPU resources. The challenge automatically scales based on CPU usage.
 
 ## Expose your challenge to the internet
-Run the following command to create a new file:
+Run the following command to make the challenge public:
 
 ```
-emacs ~/kctf-demo/demo-challenge/config/chal.conf
+sed -i s/public:\ false/public:\ true/ challenge.yaml
```
sroettger, line 126: I think it would be nicer to tell the user to edit the file. That way they will get exposed to what the challenge.yaml looks like and that it's pretty simple

docs/google-cloud.md
```diff
@@ -92,44 +90,43 @@ In this step, you'll learn how to:
...
 ```
-make start
+kctf chal start
 ```
 
 This step can take a minute. It reserves an IP address for your challenge and redirects traffic to your docker containers when someone connects to it. Wait for it to finish before continuing.
 
 While you wait, some important information to be aware of:
  * You should only expose your challenge to the internet once the challenge is ready to be released to the public. Don't expose your challenge too early or the challenge will leak.
- * The port exposed by the challenge is configured by nsjail (see `config/nsjail.cfg`). **By default, the internal nsjail port 1337 is exposed externally on port 1**.
```
sroettger, line 132: advise here that you can use kubectl chal debug port-forward for testing?

docs/google-cloud.md
```diff
@@ -10,7 +10,7 @@ The purpose of this walkthrough is to guide you through the configuration of the
 
   **Note**: If not already doing so, you can also open this walkthrough directly in [Google Cloud Shell](https://console.cloud.google.com/cloudshell/open?git_repo=https://github.com/google/kctf&tutorial=docs/google-cloud.md&shellonly=true).
```
sroettger, line 11: can we add a note here that you should install netcat in that case and replace all mentions of telnet with netcat below?

docs/google-cloud.md
```diff
@@ -160,49 +157,41 @@ In the next step we'll see how to edit the challenge, add a proof of work to pre
...
-make stop
-```
-
-To test the deletion, run:
-```
-telnet $(make ip) 1
+kctf chal stop
 ```
 
 If deletion worked correctly, the connection will fail and an error is returned instead.
```
sroettger: this is outdated since the telnet command above was deleted

docs/local-testing.md
```diff
@@ -4,258 +4,212 @@ In this walkthrough, you will learn how to use the kCTF infrastructure.
...
+### Set the right umask
 
 Since nsjail will run commands as another user, the challenge files need to be readable by all users. To enable this, set the following umask:
-```
+```bash
 umask a+rx
 ```
 
-## Testing with Docker only
+This should also be setup in our `.bashrc` and `.bash_profile` to ensure you never forget to set it.
```
sroettger: not sure if we should advise people to reduce their security. If you have a restrictive umask like that there's probably a reason

docs/local-testing.md
```diff
@@ -4,258 +4,212 @@ In this walkthrough, you will learn how to use the kCTF infrastructure.
...
 
-### Create basic demo challenge
-```
-CHALDIR=$(mktemp -d)
-kctf-setup-chal-dir $CHALDIR
-kctf-chal-create test-1
+### Enable user namespaces
+Some Linux distributions don't have user namespaces by default, to enable them in Debian or Ubuntu execute:
+```bash
+echo 'kernel.unprivileged_userns_clone=1' | sudo tee -a /etc/sysctl.d/00-local-userns.conf
```
sroettger, line 45: add a note that this has security implications?

docs/local-testing.md
```diff
@@ -4,258 +4,212 @@ In this walkthrough, you will learn how to use the kCTF infrastructure.
...
-    chmod +x ./kind
-    sudo mv ./kind /usr/local/bin/kind
-    ```
-    MacOS users can follow the same approach, but might need to adjust the path. Windows users can [follow these instructions](https://kind.sigs.k8s.io/docs/user/quick-start/).
-1. **Docker for Desktop** – Windows users can try the built-in Kubernetes cluster. See the [WSL1 instructions](#wsl1).
+### Create basic demo challenge
+To create a challenge from a skelleton you can run the following command:
+```bash
+kctf chal create chal-sample && cd chal-sample
+```
```
sroettger, line 82: mention the template param?

docs/local-testing.md
```diff
@@ -4,258 +4,212 @@ In this walkthrough, you will learn how to use the kCTF infrastructure.
...
-Open WSL 1 and enter:
-```
-echo -e '#!/bin/bash\n$(basename $0).exe $@' | tee  $HOME/.local/bin/docker $HOME/.local/bin/kubectl
-chmod +x $HOME/.local/bin/docker $HOME/.local/bin/kubectl
 ```
 
-Make sure you are running the Windows version of Docker and kubectl:
-<!-- {% raw  %} -->
+### Reading logs
+To troubleshoot unhealthy challenges, we need to read the logs of the failing healthcheck. We need to do that using `kubectl`.
```
sroettger, line 177: why not `kctf chal debug logs --container healthcheck` ?

sroettger (approved):
nice!

docs/google-cloud.md
sirdarckcat: is there a downside to telnet?

docs/google-cloud.md
sirdarckcat: they are needed, they are on the line below

docs/google-cloud.md
sirdarckcat: oh, this is a UI that pops out on the walkthrough, and it happens with the `walkthrough-enable-apis` element. the project is configured with the right thing at the beginning.

docs/google-cloud.md
sirdarckcat: hmm, done. but shouldn't we do that on `kctf config create` though?

docs/google-cloud.md
sirdarckcat: sure.

docs/local-testing.md
sirdarckcat: what are they?

docs/local-testing.md
sirdarckcat: it's necessary though. otherwise kctf should be compatible with restrictive umasks and do chmod on the Dockerfile for files that end up behind nsjail

docs/local-testing.md
sirdarckcat: it doesn't work. #262

docs/google-cloud.md
sroettger: but not if you're following it from your laptop right?

docs/google-cloud.md
sroettger: yes, it has weird behavior with newlines. I think it sends \r?
Also if I see a project talking about telnet I will assume it's from the 90s ;)

docs/local-testing.md
sroettger: but we already warn on kctf/activate. So if you forget about it, we will tell you

docs/local-testing.md
sroettger: increased kernel attack surface, that's why some distros don't enable them by default

docs/local-testing.md
sirdarckcat: I'll remove it, but `kctf activate` is a bit late. this will cause issues with gctf I think =/

docs/local-testing.md
sroettger: I see, then let's fix #262 instead

docs/google-cloud.md
sirdarckcat: ok, yea makes sense.

sirdarckcat:
@goedi02 this is ready now! :)

PR #277, opened 2021-03-29
add link to vrp code

sroettger:


sroettger:
I also replaced the nc command with socat + pty now

sirdarckcat (approved):

PR #294, opened 2021-05-27
use our own kubectl

sirdarckcat:


sroettger (approved):

sroettger:
did you grep for other cases like this?

sirdarckcat:
kindof!

PR #295, opened 2021-05-27
enumerate all capabilities in setpriv

sroettger:
Using -all in setpriv requires the libcap to be in sync with the host kernel.
Since we don't control the host kernel, we can't ship a working libcap.
Instead, we check how many caps the kernel supports (from /proc) and drop all of those.


sirdarckcat:
build didnt like it

PR #305, opened 2021-06-28
check tls validity more often

sirdarckcat:


sroettger:
I also have that in my other pull request

Issue #140, opened 2020-10-26
Support multiple kCTF clusters in the same project


sroettger:
This is desired if you want to provide challenges in different regions.

Some things that are broken currently:
- [x] subnet is hardcoded in the startup script
- [x] DNS entries get overwritten when deploying challenges with the same name
- [x] With multiple clusters I am not sure if I should be using a single kctf-conf/ directory with different cluster configs in there, or if it's ok to have multiple kctf-conf directories
- [x] Given that the cluster configs are setup using symlinks I cannot run commands in multiple clusters in different shells. Maybe we should allow changing the cluster config based on an environment variable.
- [ ] https://github.com/google/kctf/issues/199 two tasks with the same name override each other

sirdarckcat:
does this cover https://github.com/google/kctf/issues/133?

sroettger:
I think that's the first bullet point?

sirdarckcat:
I closed the other one then, feel free to reopen if you want to track each bullet independently.

sirdarckcat:
I think then this is also https://github.com/google/kctf/issues/16 - closing that one too

sirdarckcat:
Re:
> [ ] subnet is hardcoded in the startup script

That means changing `172.16.0.32/28` to something else somehow in the line below
https://github.com/google/kctf/blob/0ef4b04032848b25c324707f8b85b8fdbf0a38d5/dist/bin/kctf-cluster#L26

sirdarckcat:
Re:
> [ ] DNS entries get overwritten when deploying challenges with the same name

I think domain names are namespaced in the global (DNS) namespace, not the project one, so we can't really fix that =/

Issue #133, opened 2020-08-29
It's not possible to have multiple clusters in the same project.


aliciafmachado:
Problems with subnet and ip.

PR #306, opened 2021-06-28
IP restrictions

sroettger:
Implement IP restrictions for ingress + load balancer

The Ingress backend will now have a backend-config that points to a cloud armor security policy. In this policy, we can filter based on IP addresses.

Similarly, the LoadBalancer service now uses the LoadBalancerSourceRanges option to filter based on source IP.

The source of truth is stored in the cloud armor policy rule with priority 0. It will get applied automatically to the load balancers after a few minutes.
The operator is passed the ranges in an environment variable during startup. In order to update it, we restart the operator.


dist/bin/kctf-cluster
```diff
@@ -320,11 +320,98 @@ EOF
...
+
+  RANGES="$1"
+
+  # update the cloud armor policy
+  set_cloud_armor_policy "${RANGES}" || return
+
+  # stop the operator
+  "${KCTF_BIN}/kubectl" delete -f "${KCTF_CTF_DIR}/kctf/resources/operator.yaml" || return
+
+  start_operator_gce || return
```
sirdarckcat, line 386: what happens when someone runs this on a local cluster?

dist/bin/kctf-cluster
```diff
@@ -320,11 +320,98 @@ EOF
...
+
+function start_operator_gce {
+  get_cloud_armor_policy || return
+  RANGES=$ret
+  # restart the operator with the new range
+  "${KCTF_BIN}/yq" eval "(.spec.template.spec.containers[0].env[] | select(.name == \"ALLOWED_IPS\")).value = \"${RANGES}\"" "${KCTF_CTF_DIR}/kctf/resources/operator.yaml" | "${KCTF_BIN}/kubectl" apply -f - || return
+}
+
+function set_cloud_armor_policy {
+  gcloud compute security-policies rules update --action allow --security-policy kctf-policy --src-ip-ranges "$1" 0 || return
```
sirdarckcat: this probably needs to be namespaced (to support more than 1 cluster per project) like the other kctf stuff

dist/bin/kctf-cluster
```diff
@@ -320,11 +320,98 @@ EOF
...
+  ret=$(gcloud compute security-policies rules describe --security-policy kctf-policy 0 --format 'value[delimiter=","](match.config.srcIpRanges)')
+}
+
+function create_cloud_armor_policy {
+  if [[ -z $(gcloud compute security-policies list --filter "name=kctf-policy" --format 'get(name)') ]]; then
+    gcloud compute security-policies create kctf-policy || return
+    gcloud compute security-policies rules create --action allow --security-policy kctf-policy --src-ip-ranges "0.0.0.0/0" 0 || return
+
+    # make sure the default policy is deny
+    gcloud compute security-policies rules update --action deny-404 --security-policy kctf-policy 2147483647 || return
```
sirdarckcat: add a comment for whats the number?

dist/bin/kctf-cluster
```diff
@@ -747,12 +836,14 @@ function kctf_cluster_resize {
...
-  echo -e "  stop:   stop the cluster" >&2
-  echo -e "  resize: resize the cluster" >&2
+  echo -e "  create:            create a new cluster config" >&2
+  echo -e "  list:              list available cluster configs" >&2
+  echo -e "  load:              load an existing cluster config" >&2
+  echo -e "  start:             start the cluster" >&2
+  echo -e "  stop:              stop the cluster" >&2
+  echo -e "  resize:            resize the cluster" >&2
+  echo -e "  set-src-ip-ranges: allow a range of IPs to access the services" >&2
+  echo -e "  get-src-ip-ranges: print the current allowed IP range" >&2
```
sirdarckcat: nit: plural in name singular in text. it's ranges I guess

dist/bin/kctf-cluster
sroettger: I don't think we actually support more than one cluster per project and it makes stuff more complicated.
Do we actually want to support this?

dist/bin/kctf-cluster
sroettger: added a check for type == gce

dist/bin/kctf-cluster
sroettger: added a comment

dist/bin/kctf-cluster
sroettger: done

dist/bin/kctf-cluster
sroettger: in this case, we would need to pass the name to the operator for example

sirdarckcat: #133 fixed the issues we had with multi-cluster support. this means v1.1 doesn't support it (so it's kindof a breaking change)

sirdarckcat: oops, I meant #140

sroettger: that bug is still open though

sirdarckcat: Yes but we only have issues at the challenge name level. It's already possible to have 2 clusters with the same project, we just haven't made the challenge names namespaced too.

If we don't fix this here, then it'll be effectively impossible to have 2 clusters in one project (right now it's just not possible to have 2 tasks with the same name).

sroettger:
Isn't the ip range still hardcoded too?

sirdarckcat:
no, we have custom subnets now https://github.com/google/kctf/pull/208

sirdarckcat:
err, we have custom networks now ;P

dist/bin/kctf-cluster
sroettger: Using a unique policy name now

sirdarckcat (approved):

PR #311, opened 2021-07-01
Support for additional domains

sroettger:
Challenge authors can provide a list of domain names and kCTF will automatically generate managed certificates for them.

### DIFF
--- a/kctf-operator/deploy/crds/kctf.dev_challenges_crd.yaml
+++ b/kctf-operator/deploy/crds/kctf.dev_challenges_crd.yaml
@@ -120,6 +120,12 @@ spec:
                           - type: string
                           description: TargetPort is not optional
                           x-kubernetes-int-or-string: true
+                        domains:
+                          description: Extra domains to get certificates for. Only used for protocol HTTPS.
+                            You need set up DNS manually by adding a CNAME entry from domain to chal-web.ctf.tld.
+                          items:
+                            type: string
+                          type: array
                       required:
                       - protocol
                       - targetPort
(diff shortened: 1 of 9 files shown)


kctf-operator/pkg/controller/challenge/service/service.go
```diff
@@ -97,24 +127,18 @@ func generateIngress(domainName string, challenge *kctfv1.Challenge) *netv1beta1
...
+		ServicePort: intstr.FromInt(int(servicePort)),
+	}
 
-		ingress.Spec.Backend = &netv1beta1.IngressBackend{
-			ServiceName: challenge.Name,
-			ServicePort: intstr.FromInt(int(servicePort)),
-		}
-		// Only one https port is supported at the moment.
-		// To support more, we will need a field to specify the domain name per ingress.
-		break
```
sirdarckcat, line 117: Can you leave this comment in? So we can add this in the future maybe. As now we have domains per port so we could do it.

kctf-operator/deploy/crds/kctf.dev_challenges_crd.yaml
```diff
@@ -120,6 +120,12 @@ spec:
                           - type: string
                           description: TargetPort is not optional
                           x-kubernetes-int-or-string: true
+                        domains:
+                          description: Extra domains to get certificates for. Only used for protocol HTTPS.
+                            You need set up DNS manually by adding a CNAME entry from domain to chal-web.ctf.tld.
```
sirdarckcat: You need *to* setup

sirdarckcat (approved):

kctf-operator/pkg/controller/challenge/service/service.go
sroettger: done (added it in the other function since it makes more sense there).

PR #312, opened 2021-07-01
start zone with kctf-

sroettger:
Otherwise, domain names that start with a digit will result in an invalid zone name


sroettger:
> [E] could not register kctf.cloud subdomain

sroettger:
This is a breaking change, since you need to delete the old cloud dns zone since the name changed.

sirdarckcat (approved):

PR #317, opened 2021-07-06
Don't panic if there are non-kctf deployments

sirdarckcat:


sirdarckcat:
a better fix probably would be to have some selector that labels these as kctf's

sroettger (approved):

PR #319, opened 2021-07-06
fix challenge example

sirdarckcat:


sirdarckcat:
https://github.com/google/kctf/blob/v1/kctf-operator/samples/mychal3.yaml

sroettger (approved):

PR #345, opened 2021-10-29
MacOS compatibility

eljeffeg:
Requires the user:
`brew install coreutils gnu-getopt`

Added binaries for x64 Darwin, but arm64 (m1) could be added as well.


eljeffeg:
@googlebot I signed it!

sroettger:
Nice, thank you for the patch!

PR #346, opened 2021-11-01
Additional MacOS Compatibility

eljeffeg:
ZONE::-2 doesn't work on MacOS ZSH.  Gives an error `-2: substring expression < 0`.  ${#ZONE} should give ZONE length on Linux as well.

### DIFF
--- a/dist/bin/kctf-cluster
+++ b/dist/bin/kctf-cluster
@@ -513,12 +513,12 @@ function kctf_cluster_start_gce {
 
   EXISTING_ROUTER=$(gcloud compute routers list --filter "name=kctf-${CLUSTER_NAME}-nat-router" --format 'get(name)')
   if [ -z "${EXISTING_ROUTER}" ]; then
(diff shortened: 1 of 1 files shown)


sroettger:
looks good, thanks!

sroettger:
hmm, looks like that change broke our CI:
ERROR: (gcloud.compute.routers.nats.list) HTTPError 400: Invalid value for field 'region': 'us-east1***'. Must be a match of regex 'a-z?'

eljeffeg:
Not sure, I tried this out on linux and it produced the same result.
```
#!/bin/bash
test="us-central1a2"
echo ${test}
echo ${#test}
echo ${test::-2}
echo ${test::${#test}-2}
```
```                                                                                                       
us-central1a2
13
us-central1
us-central1
```

eljeffeg:
Oh, I know.. I copied in an extra `}` at the end.  I was working from a different directory and just did a copy paste.

sroettger:
Ahh, I totally missed that in the review :)

PR #354, opened 2022-01-10
Upgrade kCTF to use Kubernetes 1.22

sirdarckcat:
https://kubernetes.io/docs/reference/using-api/deprecation-guide/#ingress-v122


sirdarckcat:
operator-sdk depends on k8s.io/api/auditregistration/v1alpha1 which was deleted https://github.com/kubernetes/kubernetes/pull/91502 as of https://v1-19.docs.kubernetes.io/docs/setup/release/notes/ v1.19 - looks like we need to upgrade the operator to v1

sirdarckcat:
so we need the auditregistration/v1alpha1 package: https://pkg.go.dev/k8s.io/api/auditregistration/v1alpha1 which was last on v0.18.9 and the network/v1 package which was first on v0.19

sirdarckcat:
https://sdk.operatorframework.io/docs/building-operators/golang/migration/

.github/workflows/update-images.yaml
```diff
@@ -75,17 +75,24 @@ jobs:
...
       with:
         version: '319.0.0'
-        service_account_email: ${{ secrets.GCR_EMAIL }}
-        service_account_key: ${{ secrets.GCR_KEY }}
+        service_account_email: ${{ secrets.GKE_EMAIL }}
 
     - name: Configure docker to use the gcloud command-line tool as a credential helper
       if: steps.modified.outputs.modified
       run: |
+        gcloud auth list
```
sroettger: is this needed now or was this for debugging?

.github/workflows/update-images.yaml
```diff
@@ -109,11 +116,11 @@ jobs:
     - uses: actions/checkout@v2
       with:
           fetch-depth: 0
-          
+
     - id: modified
       name: Check for modified paths
       run: |
-        PATHS=(".github/workflows/update-images.yaml" "kctf-operator/*" "kctf-operator/build/**" "kctf-operator/cmd/**" "kctf-operator/pkg/**" "kctf-operator/version/**")
```
sroettger, line 116: do you remember why we didn't have "**" here before? Maybe we didn't want to trigger the workflow if some files changed?

.github/workflows/update-images.yaml
```diff
@@ -286,8 +313,12 @@ jobs:
     - name: Commit
       run: |
         # git add returns success for files that exist and haven't been modified
-        git add kctf-operator/pkg/resources/constants.go
+        git add kctf-operator/resources/constants.go
+        git add kctf-operator/operator.yaml
         git add dist/resources/operator.yaml
```
sroettger, line 314: why are there two operator.yaml?

dist/bin/kctf-cluster
```diff
@@ -398,8 +398,8 @@ function start_operator_gce {
   SUFFIX=$(echo "${PROJECT}-${CLUSTER_NAME}-${ZONE}" | sha1sum)
   POLICY_NAME="kctf-policy-${SUFFIX:0:16}"
   # restart the operator with the new range
-  "${KCTF_BIN}/yq" eval "(.spec.template.spec.containers[0].env[] | select(.name == \"ALLOWED_IPS\")).value = \"${RANGES}\"" "${KCTF_CTF_DIR}/kctf/resources/operator.yaml" \
-    | "${KCTF_BIN}/yq" eval "(.spec.template.spec.containers[0].env[] | select(.name == \"SECURITY_POLICY\")).value = \"${POLICY_NAME}\"" - \
+  "${KCTF_BIN}/yq" eval "(((.|select(has(\"spec\"))).spec|select(has(\"template\"))).template.spec.containers[0].env[] | select(.name == \"ALLOWED_IPS\")).value = \"${RANGES}\"" "${KCTF_CTF_DIR}/kctf/resources/operator.yaml" \
```
sroettger, line 401: what does this do? :)

dist/bin/kctf-cluster
```diff
@@ -431,28 +431,30 @@ function create_cloud_armor_policy {
...
+  OPERATOR_IMAGE=$("${KCTF_BIN}/yq" eval '.spec.template.spec.containers[].image | select(.=="*kctf-operator*")' "${KCTF_CTF_DIR}/kctf/resources/operator.yaml")
   if [[ $? -ne 0 ]]; then
     echo "Failed to find the operator image." >&2
     return 1
   fi
 
   # The operator needs to create some subresources, e.g. the gcsfuse service account
   for i in {1..100}; do
-    "${KCTF_BIN}/kubectl" get pods --namespace kctf-system -o=jsonpath="{.items[?(@.spec.containers[0].image==\"${OPERATOR_IMAGE}\")].status.containerStatuses[0].ready}" | grep "true" && break
+    "${KCTF_BIN}/kubectl" get pods --namespace kctf-operator-system -o=jsonpath='{.items[].status.containerStatuses[?(@.ready==true)].imageID}' | grep "${OPERATOR_IMAGE}" && break
```
sroettger, line 451: is the extra namespace needed or we just don't know where it's coming from? :)

kctf-operator/build-and-deploy-operator.sh
```diff
@@ -20,10 +20,6 @@ if [[ -z "${PROJECT}" ]]; then
   exit 1
 fi
 
-# If there's a change in the CRD, we need to regenerate it and apply it to the cluster
```
sroettger, line 23: this is not needed anymore?

kctf-operator/bundle/manifests/kctf-operator.clusterserviceversion.yaml
```diff
@@ -0,0 +1,390 @@
+apiVersion: operators.coreos.com/v1alpha1
```
sroettger, line 1: is this the same file as above?

kctf-operator/bundle/manifests/kctf.dev_challenges.yaml
```diff
@@ -1,6 +1,9 @@
 apiVersion: apiextensions.k8s.io/v1
```
sroettger, line 1: don't we have another dev_challenges file? what's the difference?

kctf-operator/config/crd/bases/kctf.dev_challenges.yaml
```diff
@@ -0,0 +1,7359 @@
+
+---
+apiVersion: apiextensions.k8s.io/v1
```
sroettger, line 3: and this is another dev_challenges file?

kctf-operator/config/manager/kustomization.yaml
```diff
@@ -0,0 +1,16 @@
...
+
+configMapGenerator:
+- files:
+  - controller_manager_config.yaml
+  name: manager-config
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+images:
+- name: controller
+  newName: eu.gcr.io/kctf-testing/kctf-operator
```
sroettger: is this supposed to be in kctf-testing?

kctf-operator/config/samples/mychal.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
```
sroettger, line 5: should we remove these mychal.yaml files?

kctf-operator/operator.yaml
```diff
@@ -0,0 +1,4549 @@
+apiVersion: v1
```
sroettger: there's another operator.yaml above, what's the difference?

sroettger (approved):

.github/workflows/update-images.yaml
sirdarckcat: ah, it was for debugging and it was useless. github redacted it xD

.github/workflows/update-images.yaml
sirdarckcat: not sure, looking at [the code](https://github.com/google/kctf/tree/3ef69474bc16039f8fcccb5f4d3341be1e8a5951/kctf-operator) it was only ignoring `deploy` and `samples`, but I think we rarely changed `samples` and `deploy`. we can make it more granular in the future if this causes delays

.github/workflows/update-images.yaml
sirdarckcat: `kctf-operator/operator.yaml` is generated by the Makefile and it is moved to `kctf-operator/operator.yaml` for release by this script. the only one that matters is `dist/resources/operator.yaml` - we could `.gitignore` the one on `kctf-operator` though, to avoid duplication

dist/bin/kctf-cluster
sirdarckcat: `operator.yaml` now has multiple yaml objects on it (before it only had one). so now we have to select the right one =(

dist/bin/kctf-cluster
sirdarckcat: it's part of the new operator framework. before operators ran on the default namespace so we manually moved it to ours. now they all get their own. I think in theory we would need to rename the operator to `kctf` instead of `kctf-operator` so it name-collides on `kctf-system` without changing the conventions.

kctf-operator/build-and-deploy-operator.sh
sirdarckcat: yea, now we always do it.

kctf-operator/bundle/manifests/kctf-operator.clusterserviceversion.yaml
sirdarckcat: yep, these are the same - `/bundle` is a generated file. we could `.gitignore`

kctf-operator/bundle/manifests/kctf.dev_challenges.yaml
sirdarckcat: its the same, it's a generated file that is moved to `/dist` by our tooling. we can `.gitignore` if helpful

kctf-operator/config/crd/bases/kctf.dev_challenges.yaml
sirdarckcat: oh, this is a base file, this one id different. here we can make changes that then get propagated up to the final one.

kctf-operator/config/manager/kustomization.yaml
sirdarckcat: hmm weird. fixed

kctf-operator/config/samples/mychal.yaml
sirdarckcat: I used them for testing, and it's the convention I think :D

kctf-operator/operator.yaml
sirdarckcat: this one could be `.gitignore`

.github/workflows/update-images.yaml
sroettger: yeah, sounds good. I just wondered if there's a chance this leads to an endless loop.
* detect changes
* rebuild
* update operator.yaml with new image tag
* detect changes
* ...

kctf-operator/config/crd/bases/kctf.dev_challenges.yaml
sirdarckcat: actually this one is generated too it seems.

dist/bin/kctf-cluster
sroettger: ah, ok. thanks :)

dist/bin/kctf-cluster
sroettger: ah ok, maybe we can follow up in the future if it's annoying us.
Do we need to change the docs anywhere?

kctf-operator/config/crd/bases/kctf.dev_challenges.yaml
sirdarckcat: https://github.com/google/kctf/blob/3a2f7ec42ee47e9d21080a3e43520c2dc62f4323/kctf-operator/Makefile#L108
https://github.com/google/kctf/blob/3a2f7ec42ee47e9d21080a3e43520c2dc62f4323/kctf-operator/Makefile#L77

looks like these are generated by different rules. we could remove them I think :grimacing:

sirdarckcat: nope we can't haha.. this broke everything

dist/bin/kctf-cluster
sirdarckcat: it seems to be hardcoded - it's based on the directory name

.github/workflows/update-images.yaml
sirdarckcat: hmm I don't think so, the steps can only run once and they wont re-run on a robot-generated change :D

dist/bin/kctf-cluster
sirdarckcat: yes, I think we need to change the docs :(

kctf-operator/operator.yaml
sirdarckcat: added to .gitignore

kctf-operator/bundle/manifests/kctf.dev_challenges.yaml
sirdarckcat: added to gitignore

kctf-operator/bundle/manifests/kctf-operator.clusterserviceversion.yaml
sirdarckcat: added to gitignore

.github/workflows/update-images.yaml
sirdarckcat: added to gitignore

PR #355, opened 2022-01-14
Set `DEBIAN_FRONTEND=noninteractive` in xss-bot

clubby789:
Prevents prompting for location which can hang forever


sirdarckcat:
ah, nice, thanks!! I incorporated it in the other PR in a slightly different way

PR #359, opened 2022-02-10
return if start_gce fails

sroettger:


sirdarckcat:
why do we wanna do that? :o

wouldn't that make us miss errors?

sroettger:
why would we continue if start_operator_gce returned an error already? It just means more unrelated errors?

sirdarckcat:
oh but then shouldn't we return the error? like `return 1` or will return preserve errno

sroettger:
From `help return`:
If N is omitted, the return status is that of the
last command executed within the function or script.

sirdarckcat:
cool!

sirdarckcat (approved):

PR #358, opened 2022-02-09
Revert "Make cloud armor optional"

sirdarckcat:
Reverts google/kctf#357

https://github.com/google/kctf/runs/5128389574?check_suite_focus=true

turns out it didnt work :(

PR #360, opened 2022-02-10
Revert "Revert "Make cloud armor optional (#357)" (#358)"

sroettger:
This reverts commit 42d7bf3e3422d1a78d80ae4a4d3ca9adcdc2c367.


sroettger:
the revert also failed the same check and now they pass for some reason..
Reverting the revert and if it fails again we can debug further (extra debug info is now enabled)

sirdarckcat (commented):
proovd

PR #365, opened 2022-03-07
add missing argument on kctf cluster resize

sirdarckcat:


dist/bin/kctf-cluster
```diff
@@ -803,7 +803,7 @@ function kctf_cluster_resize {
   fi
 
   OPTS="h"
-  LONGOPTS="help,machine-type:,min-nodes:,max-nodes:,num-nodes:,pool-name:,old-pool"
+  LONGOPTS="help,machine-type:,min-nodes:,max-nodes:,num-nodes:,pool-name:,old-pool:"
   PARSED=$(${GETOPT} --options=$OPTS --longoptions=$LONGOPTS --name "kctf chal create" -- "$@")
```
sirdarckcat: shouldnt this say `kctf cluster resize`? instead of `kctf chal create`

sroettger: ah yes, can you fix that?
We might have that problem in other places as well since I usually copy pasted those lines. The name parameter is used when printing error messages, so the bug is not too bad fortunately.

sroettger (approved):

dist/bin/kctf-cluster
sirdarckcat: fixed :)

PR #375, opened 2022-06-17
Fix two bugs in xss-bot

terjanq:
`await page.goto(url)` can take forever if a page loads infinite image for example
`page.close()` only closes one page which makes other opened tabs stay forever.


terjanq:
@sroettger

PR #379, opened 2022-06-21
Revert back to w/o timeout

terjanq:
It turns out that `{timeout: 1500}` sets a timeout for the entire request to finish but throws when it is exceeded. This error would need to get caught and then decided upon what to do which complicates things. 

Changed the order so that page.goto can be awaited but still closed after a timeout.


dist/challenge-templates/xss-bot/challenge/bot.js
```diff
@@ -6,6 +6,7 @@ const DOMAIN = process.env.DOMAIN;
 if (DOMAIN == undefined) throw 'domain undefined'
 const REGISTERED_DOMAIN = process.env.REGISTERED_DOMAIN;
 const BLOCK_SUBORIGINS = process.env.BLOCK_SUBORIGINS == "1";
+const BOT_TIMEOUT = process.env.BOT_TIMEOUT || 60;
```
sroettger: should this be 60000?

terjanq: Yes! thanks

PR #381, opened 2022-07-15
namespace kctf zones by cluster name

sirdarckcat:
This may be a breaking change on upgrade 🫣


sroettger:
bash: /home/runner/work/kctf/kctf/kctf-operator/bin/controller-gen: No such file or directory

sroettger:
No idea why the test fails. PR looks good to me though

Issue #384, opened 2022-08-20
Host-built sample challenge not guaranteed to work in challenge container


gkelly:
In the [local testing walkthrough](https://google.github.io/kctf/local-testing.html) there's an instruction to build the challenge binary using the host toolchain:

```
make -C challenge
```

However, my Ubuntu 22.04 host builds a binary with a different glibc version requirement and this results in the challenge not running in the container:

```
[I][2022-08-20T20:44:18+0000] Executing '/home/user/chal' for '[STANDALONE MODE]'
/home/user/chal: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by /home/user/chal)
```

sroettger:
Yeah, for the GoogleCTF we solved this by building all challenge binaries in a Dockerfile themselves:
https://github.com/google/google-ctf/blob/master/2021/quals/kctf/challenge-templates/pwn/challenge/Makefile

We should probably upstream that change. The original motivation for it was to allow anyone to rebuild all challenge binaries without depending on certain packages on the host.

sroettger:
@sirdarckcat wdyt?

sirdarckcat:
I feel it's a little bit weird it's in a Makefile though (I would expect the Makefile to be called within the Dockerfile, not the other way around).

We had some ideas about making all attachments "scratch" docker images iirc? Although I don't remember how we were planning to reference them

PR #385, opened 2022-08-21
Be more resilient to host/challenge glibc mismatch (#384)

gkelly:
Because the challenge runtime container and host build environment might
diverge, link the sample challenge statically to give a higher
likelihood that it can still run inside the challenge container.


sroettger:
Looks good, thank you!

PR #386, opened 2022-09-09
Support zsh without PROMPT_SUBST

### LINKED ISSUE #383: PS1 set by `activate` doesn't render properly in zsh
After running `source kctf/activate` I see:

```
[*] kCTF environment activated. Run "deactivate" to exit.
[*] To create a cluster config, run "kctf cluster create"
gkelly@pragma> kCTF[ctf=heist$(_kctf_config_string)$(_kctf_chal_string)] > 
```

Which is my original PS1 (`gkelly@pragma> `) with a partially-rendered kCTF status line. I tested this in bash with the same checkout and it works fine.

sroettger:
Fixes #383

### DIFF
--- a/dist/activate
+++ b/dist/activate
@@ -285,7 +285,12 @@ function _kctf_activate {
   else
     _kctf_log "To create a cluster config, run \"kctf cluster create\""
   fi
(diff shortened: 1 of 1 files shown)


sirdarckcat:
logic makes sense

PR #392, opened 2023-04-19
Update activate

trvon:
Moved "export KCTF_CTF_DIR" into macOS check to use grealpath instead of realpath which doesn't seem supported by default.

### DIFF
--- a/dist/activate
+++ b/dist/activate
@@ -24,13 +24,13 @@ if [[ "$OSTYPE" =~ ^darwin.* ]]; then
 
   STAT="gstat"
   MKTEMP="gmktemp"
-  if ! grealpath &> /dev/null
-  then
-    export KCTF_CTF_DIR="$(grealpath --no-symlinks "$(dirname "${BASH_SOURCE-$0}")/..")"
-  else
-    export KCTF_CTF_DIR="$(realpath --no-symlinks "$(dirname "${BASH_SOURCE-$0}")/..")"
+
+  script_dir="$(dirname "${BASH_SOURCE-$0}")"
+  if [[ "$script_dir" == "." ]]; then
+    script_dir="../."
   fi
-  
+  export KCTF_CTF_DIR="$(realpath "$(dirname "${script_dir}")")"
+  unset script_dir
 else
   KCTF_YQ_URL="https://github.com/mikefarah/yq/releases/download/v4.2.0/yq_linux_amd64"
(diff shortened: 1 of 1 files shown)


sroettger:
Hi, thanks for the pull request!
On my mac I have the opposite problem, realpath is available but grealpath is not..

Could you make it check which one exists and then put that in an environment variable maybe?

trvon:
Real path does exist on my system. I should correct my previous statement to be that how realpath is used, does not seem to be have the intended behavior. In my environment, when I source the kctf/activate after a clean download I see:

```
realpath: illegal option -- -
usage: realpath [-q] [path ...]
kctf/activate:source:50: no such file or directory: /kctf/bin/kctf-log
gstat: cannot stat '/kctf/bin/kctf-cluster': No such file or directory
_kctf_check_umask:6: command not found: _kctf_log_err
_kctf_error_cleanup:unset:12: no such hash table element: _kctf_log
_kctf_error_cleanup:unset:13: no such hash table element: _kctf_log_err
```
 I can update my commit to include the follow so it will work in both environments. 

```
  if ! grealpath &> /dev/null
  then
    export KCTF_CTF_DIR="$(grealpath --no-symlinks "$(dirname "${BASH_SOURCE-$0}")/..")"
  else
    export KCTF_CTF_DIR="$(realpath --no-symlinks "$(dirname "${BASH_SOURCE-$0}")/..")"
  fi
```

sroettger:
ah, that makes sense.
I think the better solution would be to skip the --no-symlinks in that case. I think it's there to support a symlinked `kctf` directory inside the challenge dir.

I.e. maybe something like this should work?
```bash
export KCTF_CTF_DIR="$(realpath "$(dirname "$(dirname "${BASH_SOURCE-$0}")")")"
```
In a quick test, that seems to handle all the cases of relative and absolute paths with a symlinked kctf dir

... Thinking more about it, it needs some handling of `source ./activate`. So maybe more like:
```bash
script_dir="$(dirname "${BASH_SOURCE-$0}")"
if [[ "$script_dir" == "." ]]; then
  script_dir="../."
fi
export KCTF_CTF_DIR="$(realpath "$(dirname "${script_dir}")")"
unset script_dir
```

trvon:
That solution worked as well in my initial testing. I'll update my branch to reflect the suggestion.

sroettger:
> That solution worked as well in my initial testing. I'll update my branch to reflect the suggestion.

Sounds good. Please let me know once you updated it.

trvon:
Just updated the code.

PR #393, opened 2023-05-16
Update nsjail in challenge image

happyCoder92:


badhacker0x1:
nsjail: error while loading shared libraries: libprotobuf.so.17: cannot open shared object file: No such file or directory

PR #406, opened 2024-04-18
Migrated kCTF from Google Container Registry to Artifact Registry

HlynurOskar:
Container Registry is deprecated and scheduled for shutdown. After May 15, 2024, Artifact Registry will host images for the gcr.io domain in Google Cloud projects without previous Container Registry usage. After March 18, 2025, Container Registry will be shut down.


sroettger (approved):

dist/bin/kctf-cluster
```diff
@@ -143,9 +143,10 @@ function kctf_cluster_create_usage {
 function kctf_cluster_create {
   # Default Configuration
   CLUSTER_TYPE="gce"
-  REGISTRY="eu.gcr.io"
+  REGISTRY="docker.pkg.dev"
```
sirdarckcat: Change the cmd line flag description to match the default

dist/bin/kctf-challenge
```diff
@@ -166,7 +166,7 @@ function push_image {
 
   case "${CLUSTER_TYPE}" in
     gce)
-      IMAGE_URL="${REGISTRY}/${PROJECT}/${CHALLENGE_NAME}-${IMAGE_NAME}:${IMAGE_ID}"
+      IMAGE_URL="${ARZONE}-${REGISTRY}/${PROJECT}/${CHALLENGE_NAME}/${IMAGE_NAME}:${IMAGE_ID}"
```
sirdarckcat: Maybe let's leave this as REGISTRY and just change the registry to include the region

dist/bin/kctf-cluster
sirdarckcat: Can you just leave this as europe-west4-docker.pkg.dev? So you don't need a new var

sirdarckcat (changes requested):
Please also update https://github.com/google/kctf/blob/4382f9119a873a0d4064de6137348dcde61e137b/dist/bin/kctf-completion#L98 CV

dist/bin/kctf-challenge
sirdarckcat: Oh and since you are changing this, can you include the cluster name on the url? So we don't collide on a project with two challenges named the same

dist/bin/kctf-challenge
```diff
@@ -166,7 +166,7 @@ function push_image {
 
   case "${CLUSTER_TYPE}" in
     gce)
-      IMAGE_URL="${REGISTRY}/${PROJECT}/${CHALLENGE_NAME}-${IMAGE_NAME}:${IMAGE_ID}"
+      IMAGE_URL="${REGISTRY}/${PROJECT}/${CLUSTER_NAME}-${CHALLENGE_NAME}/${IMAGE_NAME}:${IMAGE_ID}"
```
sirdarckcat, line 169: Is there a difference between slashes here? Maybe just add slashes too, for consistency.

dist/bin/kctf-completion
```diff
@@ -95,7 +95,7 @@ function _kctf_complete_chal() {
...
-GCP_REGISTRIES="gcr.io asia.gcr.io eu.gcr.io us.gcr.io"
+GCP_REGISTRIES="pkg.dev docker.pkg.dev"
```
sirdarckcat, line 98: This should include the zones I think

sirdarckcat (changes requested):

dist/bin/kctf-challenge
HlynurOskar: There is a significance to the slashes, it's a path that gets parsed.
Although, some want the path to be more like: 
```IMAGE_URL="${REGISTRY}/${PROJECT}/challenges/${CLUSTER_NAME}-${CHALLENGE_NAME}-${IMAGE_NAME}:${IMAGE_ID}"```

sirdarckcat: What if we have a dash in the cluster? Wouldn't that cause a collision?

pl4nty: the registry should be shared by all challenges, so we can create it in `kctf cluster create` rather than managing several in `kctf challenge`. I used `IMAGE_URL="${REGISTRY}/${PROJECT}/${CLUSTER_NAME}/${CHALLENGE_NAME}-${IMAGE_NAME}:${IMAGE_ID}"`

should registry creation be part of this PR? Container Registry is shutting down very soon

PR #409, opened 2025-03-21
Image update: ubuntu to 24.04, gcsfuse to 1.x, nsjail to latest

implr:
Bumping all the (very) old dependencies our containers have.
To be honest I'm not sure how to test it other than trying to deploy it to a cluster (after it builds&releases). The images build fine, but eg. something nonobvious could have broken in gcsfuse.


sroettger:
nice!

PR #410, opened 2025-03-21
Update update-images.yaml

sroettger:
The action fails after ubuntu update, saying that 3.7 is not available. https://github.com/google/kctf/actions/runs/13995620258/job/39189928063#step:6:334

The newest non-alpha version listed in https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json available on Ubuntu 24.04 is 3.13.2

Also update setup-gcloud to v2 and use the latest version

### DIFF
--- a/.github/workflows/update-images.yaml
+++ b/.github/workflows/update-images.yaml
@@ -181,7 +181,7 @@ jobs:
     - name: 'Setup go version necessary for operator'
       uses: actions/setup-go@v5
       with:
-        go-version: 'latest'
+        go-version: '1.24.1'
 
     - name: Build image
       if: steps.modified.outputs.modified
--- a/.github/workflows/update-images.yaml
+++ b/.github/workflows/update-images.yaml
@@ -89,9 +89,9 @@ jobs:
         credentials_json: '${{ secrets.GKE_KEY }}'
 
     - name: 'Set up Cloud SDK'
-      uses: 'google-github-actions/setup-gcloud@v0'
+      uses: 'google-github-actions/setup-gcloud@v2'
       with:
-        version: '319.0.0'
+        version: 'latest'
         service_account_email: ${{ secrets.GKE_EMAIL }}
 
     - name: Configure docker to use the gcloud command-line tool as a credential helper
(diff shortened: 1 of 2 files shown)


sroettger:
@implr @sirdarckcat 
I don't know what I'm doing, does this change make sense? :)

sroettger:
```
/home/runner/work/kctf/kctf/kctf-operator/bin/controller-gen rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x9efdfe]
```

/tableflip

.github/workflows/update-images.yaml
```diff
@@ -89,9 +89,9 @@ jobs:
         credentials_json: '${{ secrets.GKE_KEY }}'
 
     - name: 'Set up Cloud SDK'
-      uses: 'google-github-actions/setup-gcloud@v0'
+      uses: 'google-github-actions/setup-gcloud@v2'
       with:
-        version: '319.0.0'
+        version: 'latest'
```
implr, line 94: I think we should pin this, otherwise it'll break unexpectedly - gcloud follows latest, python/ubuntu is stuck on old

.github/workflows/update-images.yaml
```diff
@@ -162,15 +162,15 @@ jobs:
...
 
     - name: Export gcloud related env variable
       run: export CLOUDSDK_PYTHON="/usr/bin/python3"
     
     - name: 'Set up Cloud SDK'
-      uses: 'google-github-actions/setup-gcloud@v0'
+      uses: 'google-github-actions/setup-gcloud@v2'
       with:
-        version: '319.0.0'
+        version: 'latest'
```
implr, line 173: ditto

implr (changes requested):
Thanks, looks good, I'm not familiar with the operator-sdk stuff though.

implr:
Replaced by #415.

PR #415, opened 2025-03-29
Large batch dependency update

implr:
Updated operator-framework, k8s, controller-runtime, gke-managed-certs, golang, builder ubuntu images and so many other transitive things. Regenerated the CRD and diffed other yamls and the Makefile with operator-framework's current scaffolding.

Supersedes #412 and #410.


implr:
OK, finally. At least it seems to work now. I'm not very confident in this though.
@sroettger @sirdarckcat do you want to take a look? Reviewing this monstrous change carefully is probably out of the question, but maybe you can skim and see if I did something obviously stupid.
Note that `kctf-operator/config/crd/bases/kctf.dev_challenges.yaml` was autoregenerated, you can skip it.

sirdarckcat:
Oh yikes, you had to update the operator?

dist/challenge-templates/pwn/challenge/Dockerfile
```diff
@@ -13,12 +13,13 @@
 # limitations under the License.
 FROM ubuntu:24.04 as chroot
 
-RUN /usr/sbin/useradd --no-create-home -u 1000 user
```
sirdarckcat, line 16: Why?

dist/challenge-templates/web/challenge/Dockerfile
```diff
@@ -13,17 +13,16 @@
 # limitations under the License.
 FROM ubuntu:24.04 as chroot
 
-RUN /usr/sbin/useradd -u 1000 user
```
sirdarckcat, line 16: Why?

dist/challenge-templates/xss-bot/challenge/bot.js
```diff
@@ -50,7 +51,7 @@ if (BLOCK_SUBORIGINS) {
     socket.state = 'LOADED';
     let cookie = JSON.parse(fs.readFileSync('/home/user/cookie'));
 
-    const context = await browser.createIncognitoBrowserContext();
+    const context = await browser.createBrowserContext();
```
sirdarckcat, line 54: Incognito is to avoid saving state. Is this not needed?

kctf-operator/Makefile
```diff
@@ -68,92 +88,189 @@ all: build
...
+	$(KUSTOMIZE) build config/crd | $(KUBECTL) delete --ignore-not-found=$(ignore-not-found) -f -
 
+.PHONY: deploy
 deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
-	$(KUSTOMIZE) build config/default | kubectl apply -f -
-
-operator.yaml: manifests kustomize ## Output controller configuration
-	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
-	$(KUSTOMIZE) build config/default > operator.yaml
```
sirdarckcat, line 119: How does one generate the operator.yaml now?

sirdarckcat (approved):
Just leaving a few q

dist/challenge-templates/xss-bot/challenge/bot.js
sroettger: This is replaced by --incognito in the args above. It's missing though if the BLOCK_SUBORIGINS is false

dist/challenge-templates/web/challenge/Dockerfile
implr: same as above

dist/challenge-templates/pwn/challenge/Dockerfile
implr: New ubuntu includes the `ubuntu` user by default (what I meant in the comment), which has uid 1000, which makes it taken, so a `useradd -u 1000` will fail, so we have to remove `ubuntu` first.
Alternatively it could stay and we'd run the challenge as `ubuntu` instead of `user`, but I didn't want to ensure nothing depended on the username.

kctf-operator/Makefile
implr: Good question. The changes in this Makefile come from generating a new scaffold with `operator-sdk init` in a separate directory and applying most of the diff. The new scaffold Makefile doesn't mention operator.yaml at all.
I'll read through operator-sdk's changelog again to see if they mentioned this.

implr: After some reading it seems that in our setup we don't need any of the stuff in bundle/ and should instead be using the output of build-installer (which you'll note has a very similar make rule to operator.yaml).

dist/challenge-templates/xss-bot/challenge/bot.js
implr: Changed to a push() outside of the if.

implr:
> Oh yikes, you had to update the operator?

I started with this assumption as that's what @sroettger did in the previous PR. Maybe it wasn't strictly necessary, but for continued existence of kctf it had to be done at some point. k8s moves pretty fast, targetting a three year old api will go wrong eventually.

kctf-operator/Makefile
implr: Pushed another big change which gets rid of bundles and uses an unified install.yaml (pretty much same as the old operator.yaml). I have a question about that though - how did you manage the operator image sha256 in operator.yaml previously? Seems like a chicken-and-egg problem, where this PR needs to land so we know the hash after actions build it, before it can be changed in the final yaml?
BTW, it seems like at current HEAD, kctf-operator/operator.yaml and dist/resources/operator.yaml have different hashes.

sroettger: This is happening in the github workflow:
https://github.com/google/kctf/blob/v1/.github/workflows/update-images.yaml#L249

which creates commits like: https://github.com/google/kctf/commit/818e3b6373ff15352987865a54a8bc5b5da0768d

implr: Oh, I didn't see that part, I thought it just fixed the challenge templates. OK then, I'll fix that to use install.yaml and it should be good.

PR #418, opened 2025-06-05
add support for nested virt

sirdarckcat:


sroettger:
Nice! Do you want to add a sample challenge (which would also give this test coverage)?
But in any case LGTM

sirdarckcat:
I ran the instructions here:
https://github.com/kata-containers/kata-containers/blob/main/tools/packaging/kata-deploy/README.md

and then added this to challenge.yaml
```
 podTemplate:
    template:
      spec:
        runtimeClassName: kata-qemu
        containers:
          - name: challenge
```

and now the challenge runs in a vm.

---

for reference, here is the complete challenge.yaml:
```yaml
apiVersion: kctf.dev/v1
kind: Challenge
metadata:
  name: demo-challenge
spec:
  deployed: true
  powDifficultySeconds: 0
  network:
    public: true
  healthcheck:
    # TIP: disable the healthcheck during development
    enabled: true
    image: eu.gcr.io/kctf-testing/demo-challenge-healthcheck:09f66401a712a419ef0b4987d9d3ed93194213c792ff80adff7abbb6d5490fa5
  image: eu.gcr.io/kctf-testing/demo-challenge-challenge:df97c62f3d784b05125c44ab310aeca054041831b5d65a05ac31f5f61703ff5f
  podTemplate:
    template:
      spec:
        runtimeClassName: kata-qemu
        containers:
          - name: challenge
```

sirdarckcat:
@sroettger wdyt should we make kctf always run the challenge container on a kata container (a vm)?

sroettger:
Is there anything that would break with this? Does it change the performance significantly, maybe startup time?

sirdarckcat:
hmm yea, I think it adds some time for kata to be installed on the node (not a lot), and there's ofc some overhead with having another virtualization layer for every vmcall/interrupt/etc but that'll depend on the workload. there will also be more RAM needed per challenge instance as now every challenge instance will need to run a kernel and all that.

maybe let's leave it as opt-in, but maybe a feature on the CRD? so you dont have to modify the podTemplate but just say `vm: true` or something like that?

sroettger:
Yeah, I think it's cool and we should add a toggle for it in the challenge.yaml.
I think it would also be fine to make it on by default, with the option to disable, since it brings much stronger isolation guarantees.

sirdarckcat:
lets see if we can see the cost it would have on gctf. @implr wdyt

sirdarckcat:
worked, one line needs to be deleted otherwise works ok.

https://github.com/kata-containers/kata-containers/blob/8c1f6e827d91b4b5c1b01440c02882249556c8cb/tools/packaging/kata-deploy/kata-deploy/base/kata-deploy.yaml#L25

PR #420, opened 2025-06-28
feat(pow): Add progress bar

N3rdL0rd:
This stems out of an issue I had playing Google CTF - I was running the PoW script as intended, but it looked like it was hanging. I quickly scripted this together and I can only hope it benefits others. Tested and functional.


sirdarckcat:
could you please just put this inside a function (to avoid code duplication) and only pass it if an argument is passed to solve_challenge (just pass it when cmd=="solve") so it runs when executed interactively but not when done programatically (like in a healthcheck)

also - CLA :D

Issue #378, opened 2022-06-21
ARM64 support


ebanDev:
Hey !
I'd like to "local test" kctf but my machine is running ARM64 and kctf doesn't support it :/ Is it planned ?
Thanks !
Eban

sroettger:
Hey there,
We're currently not working on this. Do you happen to know what's required
for ARM support?
One thing I can think of is the binaries that we're using for kind etc:
https://github.com/google/kctf/blob/v1/dist/activate#L16
And then I guess we would have to build our docker images for arm64 too.

PR #425, opened 2025-09-11
activate: download arm64 Linux versions of binaries

zhuowei:
Detect arm64 Linux and download jq, kind, and kubectl for arm64.

Per #378, this isn't enough for full arm64 support (the containers would also need to be rebuilt), but this is at least useful for running kctf while using qemu-user emulation for the containers.

### DIFF
--- a/dist/activate
+++ b/dist/activate
@@ -31,6 +31,19 @@ if [[ "$OSTYPE" =~ ^darwin.* ]]; then
   fi
   export KCTF_CTF_DIR="$(realpath "$(dirname "${script_dir}")")"
   unset script_dir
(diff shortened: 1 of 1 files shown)


dist/activate
```diff
@@ -31,6 +31,19 @@ if [[ "$OSTYPE" =~ ^darwin.* ]]; then
   fi
   export KCTF_CTF_DIR="$(realpath "$(dirname "${script_dir}")")"
   unset script_dir
+elif [[ "$HOSTTYPE" =~ ^aarch64.* ]]; then
```
sroettger, line 34: Can you add a log warning somewhere that aarch64 is not fully supported (with what you mentioned in the pull request)?
Otherwise lgtm!

